PR #36, opened 2026-02-23
feat(frontend): add live metrics freshness controls

TheArchitectit:
## Summary
- Add stream freshness controls to Live Metrics: manual refresh, configurable polling interval, and delayed/stale stream signaling.
- Add a focused "Top Providers (Live Window)" table so empty windows and low traffic can be diagnosed without leaving the page.
- Keep scope intentionally narrow as PR chunk 2 for the Live Metrics rollout.

## Changes Made
- `packages/frontend/src/pages/LiveMetrics.tsx`
  - Add manual `Refresh Now` action and `5s/10s/30s` polling controls.
  - Add stale detection based on last update age and selected poll interval.
  - Add visibility-aware polling behavior (resume refresh when tab is focused).
  - Add provider summary table (top 6 by live-window request volume) with success rate, avg latency, and cost.

## Testing
- `bun run typecheck` (frontend package) ✅
- `bun run build` (frontend package) ✅
- `bun test` (frontend package) ✅
- `lsp_diagnostics` for modified file ✅

## Scope / Notes
- Single-file frontend-only change.
- No backend/API contract changes.
- Diff size: 155 lines changed (148 insertions, 7 deletions).

PR #38, opened 2026-02-23
feat(frontend): add live velocity and pulse sections

TheArchitectit:
## Summary
- Add `Request Velocity (Last 5 Minutes)` to highlight minute-over-minute request movement.
- Add `Provider Pulse (5m)` and `Model Pulse (5m)` sections so operators can quickly identify active traffic sources in the live window.
- Keep this as a focused phase-3 Live Metrics chunk.

## Changes Made
- `packages/frontend/src/pages/LiveMetrics.tsx`
  - Add `velocitySeries` derived state and line chart card.
  - Add provider/model pulse aggregations and list rendering.
  - Add provider pulse bar chart for quick top-provider comparison.

## Testing
- `bun run typecheck` (frontend package) ✅
- `bun run build` (frontend package) ✅
- `bun test` (frontend package) ✅
- `lsp_diagnostics` on modified file ✅

## Scope / Notes
- Single-file frontend-only change.
- No backend/API contract changes.
- Diff size for this commit: 216 insertions (under the 500-line cap).
- This PR is stacked on #36 and will be kept draft until prior chunks merge.


packages/frontend/src/pages/LiveMetrics.tsx
```diff
@@ -151,6 +200,133 @@ export const LiveMetrics = () => {
...
+        return { time: bucket.time, velocity: bucket.requests };
+      }
+
+      const prev = arr[index - 1];
+      return {
+        time: bucket.time,
+        velocity: bucket.requests - prev.requests,
+      };
+    });
+  }, [minuteSeries]);
```
mcowger, line 277: **Bug: `velocitySeries` computes wrong delta**

`minuteSeries` buckets already hold the **absolute request count for that individual minute** — they are not a running cumulative total. So `bucket.requests - prev.requests` doesn't produce a meaningful velocity delta; in real traffic you'll get values near zero or even negative whenever the previous bucket happened to have more requests.

If the goal is "requests per minute over the live window", `bucket.requests` *is* already that value — just use it directly:

```ts
const velocitySeries = useMemo(() => {
  return minuteSeries.map((bucket) => ({
    time: bucket.time,
    velocity: bucket.requests,
  }));
}, [minuteSeries]);
```

If you genuinely want a second-derivative acceleration signal, document that clearly and make sure `minuteSeries` is a cumulative series first.

packages/frontend/src/pages/LiveMetrics.tsx
```diff
@@ -49,19 +70,47 @@ export const LiveMetrics = () => {
...
+    if (!isVisible) {
+      return;
+    }
+
+    const interval = setInterval(() => {
+      void loadData(true);
+    }, pollIntervalMs);
+
     return () => clearInterval(interval);
+  }, [isVisible, pollIntervalMs]);
```
mcowger: **Bug: stale closure — `loadData` missing from dependency arrays**

`loadData` is defined as a plain `async` function in the component body. Both this effect and the visibility effect below capture the version of `loadData` that existed at mount time. While it doesn't cause a bug *today* (the function only closes over `setters`), this is fragile: any future state/prop dependency added to `loadData` will silently use stale values.

The clean fix is `useCallback`:

```ts
const loadData = useCallback(async (silent = false) => {
  // ...existing body...
}, []); // no external deps
```

Then add `loadData` to both effect dependency arrays:
```ts
}, [isVisible, pollIntervalMs, loadData]);
// and
}, [loadData]);
```

packages/frontend/src/pages/LiveMetrics.tsx
```diff
@@ -151,6 +200,133 @@ export const LiveMetrics = () => {
...
+
+    return Array.from(rows.entries())
+      .map(([label, row]) => ({
+        label,
+        requests: row.requests,
+        successRate: row.requests > 0 ? (row.success / row.requests) * 100 : 0,
+      }))
+      .sort((a, b) => b.requests - a.requests)
+      .slice(0, 8);
+  }, [liveRequests]);
```
mcowger: **Redundant aggregation: `providerRows` and `providerPulseRows` iterate `liveRequests` twice**

Both `providerRows` (line 205) and `providerPulseRows` (here) loop over `liveRequests` and compute the same `requests` and `successRate` per-provider. The only differences are that `providerRows` additionally tracks `avgLatency` and `totalCost`, and caps at 6 vs 8.

Merge into one `useMemo` and project/slice at the call site:

```ts
const providerRows = useMemo(() => {
  const providers = new Map<string, { requests: number; success: number; totalLatency: number; totalCost: number }>();
  for (const request of liveRequests) {
    const provider = request.provider || 'unknown';
    const row = providers.get(provider) ?? { requests: 0, success: 0, totalLatency: 0, totalCost: 0 };
    row.requests += 1;
    if ((request.responseStatus || '').toLowerCase() === 'success') row.success += 1;
    row.totalLatency += Number(request.durationMs || 0);
    row.totalCost += Number(request.costTotal || 0);
    providers.set(provider, row);
  }
  return Array.from(providers.entries())
    .map(([provider, row]) => ({
      provider,
      requests: row.requests,
      successRate: row.requests > 0 ? (row.success / row.requests) * 100 : 0,
      avgLatency: row.requests > 0 ? row.totalLatency / row.requests : 0,
      totalCost: row.totalCost,
    }))
    .sort((a, b) => b.requests - a.requests);
}, [liveRequests]);

// Then at call sites:
// Top Providers card: providerRows.slice(0, 6)
// Provider Pulse chart: providerRows.slice(0, 6)
// Provider Pulse Details list: providerRows.slice(0, 8)
```

packages/frontend/src/pages/LiveMetrics.tsx
```diff
@@ -151,6 +200,133 @@ export const LiveMetrics = () => {
...
+
+  const renderPulseList = (rows: PulseRow[], emptyText: string) => {
```
mcowger: **`renderPulseList` should be a proper React component, not a plain function**

Defining JSX-returning functions inside a component body means React cannot diff them across renders — every call creates a new React tree from scratch rather than reconciling. This causes unnecessary unmount/remount of all child elements on every parent re-render.

Extract it as a named component:

```ts
const PulseList: React.FC<{ rows: PulseRow[]; emptyText: string }> = ({ rows, emptyText }) => {
  if (rows.length === 0) {
    return <div className="text-text-secondary text-sm py-2">{emptyText}</div>;
  }
  return (
    <div className="space-y-2 max-h-[320px] overflow-y-auto pr-1">
      {rows.map((row) => (
        // ...existing markup...
      ))}
    </div>
  );
};
```

Place it outside (or above) the `LiveMetrics` component, and use `<PulseList rows={...} emptyText={...} />` at the call sites.

packages/frontend/src/pages/LiveMetrics.tsx
```diff
@@ -321,6 +528,156 @@ export const LiveMetrics = () => {
...
+        >
+          {providerPulseRows.length === 0 ? (
+            <div className="h-56 flex items-center justify-center text-text-secondary">
+              No provider traffic in the selected live window.
+            </div>
+          ) : (
+            <div className="h-56">
+              <ResponsiveContainer width="100%" height="100%">
+                <BarChart
+                  data={providerPulseRows.slice(0, 6)}
```
mcowger: **Mismatch: chart slices to 6 but the memo caps at 8**

`providerPulseRows` is memoized with `.slice(0, 8)`, but here the chart slices it again to 6. The "Provider Pulse Details" list below uses the full 8-entry array, so a user can see entries in the list that don't appear in the chart — confusing.

Pick one number and be consistent. If the chart fits 6 bars comfortably, cap the memo at 6 and remove the `.slice(0, 6)` here:

```ts
.slice(0, 6); // in the useMemo
// ...
data={providerPulseRows} // no extra slice needed
```

Or keep 8 in the memo and slice to 6 in both the chart and the details list.

packages/frontend/src/pages/LiveMetrics.tsx
```diff
@@ -321,6 +528,156 @@ export const LiveMetrics = () => {
...
+                </div>
+              ))}
+            </div>
+          )}
+        </Card>
+      </div>
+
+      <div
+        className="grid gap-4 mb-4 flex-col lg:flex-row"
+        style={{ gridTemplateColumns: '1fr 1fr' }}
```
mcowger: **Nit: use `grid-cols-2` instead of an inline style for symmetric grids**

For the symmetric `1fr 1fr` case a Tailwind class is cleaner and avoids the inline style:

```tsx
<div className="grid grid-cols-2 gap-4 mb-4">
```

The inline `style` is already justified for the asymmetric `1.2fr 1fr` grid lower in the file — no need to carry the pattern here where a utility class suffices. Same applies to the second `1fr 1fr` grid at line 664.

PR #65, opened 2026-02-27
fix(concurrency): show actual in-flight requests

TheArchitectit:
## Summary

Fix the `/v0/management/concurrency` endpoint to show actual **concurrent/in-flight requests** instead of total requests per time bucket.

### Problem
The original query counted all requests in minute buckets (requests per minute), which is a throughput metric, not a concurrency metric. Users expected to see currently active requests.

### Solution
Changed query to filter for requests where `duration_ms IS NULL`:
- `duration_ms IS NULL` indicates request started but not yet completed
- Added 1-hour window filter to avoid stale data from orphaned records
- Returns current active request count grouped by provider and model

### Changes
- `routes/management/usage.ts`: Rewrote concurrency endpoint query
  - Removed time bucket grouping (was causing misleading "historical" view)
  - Added `duration_ms IS NULL` filter for in-flight requests
  - Added 1-hour start time window for data freshness

**Note**: This PR depends on #64 (LiveTab DnD Integration) and should be merged after it.

## Test plan

- [ ] Concurrency card shows actual in-flight request count
- [ ] When no requests are active, shows "No active concurrent requests"
- [ ] Provider breakdown shows correct counts per provider
- [ ] Data refreshes every 5 seconds
- [ ] Old/completed requests don't appear in the count

Lines changed: 30 insertions (+), under 500 line limit ✓

### DIFF
--- a/packages/backend/src/routes/management/usage.ts
+++ b/packages/backend/src/routes/management/usage.ts
@@ -350,4 +350,34 @@ export async function registerUsageRoutes(
       }
     }
   });
+
+  fastify.get('/v0/management/concurrency', async (_request, reply) => {
+    try {
+      const db = usageStorage.getDb();
+      const schema = getSchema();
+
+      // Query currently in-flight requests (started but not yet completed)
+      // duration_ms IS NULL indicates request is still active
+      const results = await db
+        .select({
+          provider: schema.requestUsage.provider,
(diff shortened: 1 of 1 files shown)


TheArchitectit:
Will base on main. Needs more work.

PR #134, opened 2026-03-18
feat: add server-side proxy for model fetching to bypass CORS

EvairXCV:
## Summary
- Adds `/v0/management/providers/fetch-models` endpoint for server-side model fetching
- Fixes browser CORS failures when fetching provider models directly
- Includes SSRF guards (localhost + cloud metadata IP block) and redirect manual handling
- Updates Providers UI to use backend proxy instead of direct fetch


mcowger:
This is a good idea - but curiously, which providers enforce CORS on their model endpoints?

PR #139, opened 2026-03-20
feat: add AES-256-GCM encryption at rest for sensitive database fields

darkspadez:
Encrypt API key secrets, OAuth tokens, provider API keys, and MCP server
headers in the database using AES-256-GCM. Encryption is controlled by the
ENCRYPTION_KEY env var and is backward-compatible (plaintext mode when unset).

- New encryption utility (src/utils/encryption.ts) with encrypt/decrypt/hash
- Add secret_hash column to api_keys for hash-based lookups
- Encrypt on write / decrypt on read in ConfigRepository
- Data migration encrypts existing plaintext on first startup with key set
- 29 passing tests covering utility and migration

### DIFF
--- a/packages/backend/src/db/config-repository.ts
+++ b/packages/backend/src/db/config-repository.ts
@@ -46,18 +46,14 @@ function toJson(value: unknown): string | unknown {
 }
 
 /**
- * Encrypt a JSON value for storage. For SQLite (text columns), encrypts the JSON string.
- * For PG (jsonb columns), stores the encrypted string as a JSON string value.
+ * Encrypt a JSON value for storage in a TEXT column.
+ * JSON-serializes the value, then encrypts the resulting string.
+ * If encryption is disabled, returns the JSON string as-is.
  */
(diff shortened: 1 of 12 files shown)


packages/backend/src/db/config-repository.ts
```diff
@@ -499,19 +546,31 @@ export class ConfigRepository {
...
+    // Try hash-based lookup first (works after encryption migration)
+    let rows = await this.db()
       .select()
       .from(schema.apiKeys)
-      .where(eq(schema.apiKeys.secret, secret))
+      .where(eq(schema.apiKeys.secretHash, hash))
       .limit(1);
 
+    // Fallback to plaintext lookup for backward compatibility (before migration)
+    if (rows.length === 0) {
```
mcowger, line 555: this should emit a ERROR level log.

mcowger:
I'm going to merge this as is to unblock, but please follow up on the rekey utility.

PR #149, opened 2026-04-03
feat: add per-model extraBody support

zicochaos:
## Summary

Adds the ability to configure `extraBody` parameters at the individual model level within a provider, not just at the provider level. When both provider-level and model-level `extraBody` are set, model-level overrides provider-level.

## Motivation

Currently, `extraBody` is only configurable at the provider level. This means all models under a provider share the same extra parameters. In practice, different models on the same provider often need different configurations.

**Example use case:** On OpenRouter, you may want to disable reasoning/thinking for Qwen models (`"reasoning": {"effort": "none"}`) while keeping it enabled for other models on the same provider. Without per-model `extraBody`, this requires creating separate providers pointing to the same endpoint — a workaround that clutters the configuration.

## Changes

**Backend:**
- **Schema** (`drizzle/schema/sqlite/provider-models.ts`, `drizzle/schema/postgres/provider-models.ts`): Added `extra_body` column to `provider_models` table
- **Config** (`config.ts`): Added `extraBody: z.record(z.any()).optional()` to `ModelProviderConfigSchema`
- **Repository** (`config-repository.ts`): Save/load `extraBody` when persisting/reading model configs from DB
- **Dispatcher** (`dispatcher.ts`): Merge model-level `extraBody` after provider-level `extraBody` (model overrides provider)

**Frontend:**
- **UI** (`Providers.tsx`): Per-model "Extra Body Fields" collapsible section with add/edit/remove key-value pairs, matching the existing provider-level Extra Body Fields UI pattern

## Migration

- SQLite: `ALTER TABLE provider_models ADD COLUMN extra_body TEXT;`
- PostgreSQL: `ALTER TABLE provider_models ADD COLUMN extra_body JSONB;`

The existing drizzle migration system should handle this automatically on startup.

## Testing

Tested with:
- OpenRouter + Qwen model: `{"reasoning": {"effort": "none"}}` → 0 reasoning tokens (vs 1400+ without)
- Local vLLM + Qwen model: `{"chat_template_kwargs": {"enable_thinking": false}}` → thinking correctly overridden

7 files changed, 137 insertions(+), 8 deletions(-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,4 +54,4 @@ packages/backend/config/
 research_and_plans/
 packages/logs.txt
 .vscode/mcp.json
-.pi/readcache/plexus-custom
+.pi/readcache/
--- a/packages/backend/src/db/config-repository.ts
+++ b/packages/backend/src/db/config-repository.ts
@@ -271,10 +271,10 @@ export class ConfigRepository {
           providerId,
           modelName: name,
           pricingConfig: toJson(cfg.pricing),
-            modelType: cfg.type ?? null,
-            accessVia: cfg.access_via ? toJson(cfg.access_via) : null,
-            extraBody: cfg.extraBody ? toJson(cfg.extraBody) : null,
-            sortOrder: idx,
+          modelType: cfg.type ?? null,
+          accessVia: cfg.access_via ? toJson(cfg.access_via) : null,
+          extraBody: cfg.extraBody ? toJson(cfg.extraBody) : null,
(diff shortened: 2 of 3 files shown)


zicochaos:
Thanks for the review! Good catch — swapped both occurrences of `JSON.stringify(cfg.extraBody)` to `toJson(cfg.extraBody)` (lines 217 and 276 in config-repository.ts). Should be consistent with the other JSON fields now.

PR #154, opened 2026-04-09
feat: Support the new Synthetic quota system

### LINKED ISSUE #150: [Feature Request] Display free tool calls time until reset for synthetic
It would be nice to display free tool calls time until reset for synthetic

Image

Image

Image

fredizzimo:
Also improve the quota card display to show more information. The quota graphs support both the new and old system, so historical data can be viewed.

image
image

fixes: https://github.com/mcowger/plexus/issues/150


fredizzimo:
I made a  couple of small changes. I renamed `toolcalls` to `toolcalls (old)`. And I also removed the `5 hour` and `toolcall` windows from new fetches, so that they don't show up in the graph, unless old data is present.

Issue #16, opened 2026-01-06
README needs update


mcowger:
The readme doesn't include the recent changes around multi end point configuration for providers, nor api priority.

Make some pretty massive improvements on Plexus over the break. 
0. Implemented pessimistic stream conversion - if the incoming API is messages, and the outgoing API is messages, plexus will not perform any transformations on the stream to maximize compatibility and performance.  
1. Switched from Hono to Fastify because Hono’s stream handling sucks.   No more API Error: terminated issues 
2. Implemented 100% coverage for all but 1 config option in the UI (excluded the admin key as I feel that should require more effort). You can configure providers, models, routing preferences, etc all in the UI, but written back to the config YAML.  
3. Implemented multi-endpoint support for providers.   So for great providers like synthetic that offer both chat completions and messages endpoints, you can declare them both, as well as which models can be used via which interfaces.  Plexus will handle the authentication data rewriting.  This pairs with…
4. By default, plexus will prioritize the selector you set to choose between providers (random, latency, throughput or cost).  But…you can specify api_priority instead, and it will first filter for providers / targets that support your incoming api type.   So if you are using Claude code and the messages endpoint, it will search first for providers that can natively support that, and THEN run the selector algorithm.   This allows for higher fidelity, possibly at the loss of some options for latency or cost reduction.   But you get to pick!
5. Rewrote the OAI -> anthropic stream conversion with a state machine.   Can now run hundreds of back to back requests from Claude code to openrouter or synthetic without a parse issue.  
6. Added support for images in anthropic protocol.

Issue #10, opened 2025-12-27
Support Responses API


mcowger:

mcowger:
Basics are working, but need to figure out what it means to implement the state tracking for responses when the backend is not responses.

PR #156, opened 2026-04-11
Improve calculations and display for energy usage.

alexispurslane:
- Revised the inference energy estimator to accept the model architecture and gpu as arguments.
- Added an accordion form to the model alias modal that lets the user:
  - fetch model architecture, size, and default dtype information from HuggingFace (safetensors + config.json info) and estimates active parameters and total parameters from that
  - edit those values (dtype gets a dropdown)
  - save those values
  - fetch those values when previously stored and edit them again
- Added a dropdown to a provider to choose the default type of GPU they have. Also saved
- Recalculates the energy usage of all past requests when a model's architecture is updated.
- Shows the amount of time in seconds you could've streamed Netflix for the same amount of energy as your prompts for the give hour/day/week/month on the Usage page
- Also compares energy usage rate between your model usage and Netflix
- Also added an energy usage over time time series graph to the usage page.


packages/backend/src/config.ts
```diff
@@ -347,6 +347,12 @@ export const ProviderConfigSchema = z
     estimateTokens: z.boolean().optional().default(false),
     useClaudeMasking: z.boolean().optional().default(false),
     quota_checker: ProviderQuotaCheckerSchema.optional(),
+    // GPU Profile settings for inference energy calculation
+    gpu_profile: z.enum(['H100', 'H200', 'GH100', 'GH200', 'B200', 'B300', 'custom']).optional(),
```
mcowger: 🔴 **Blocker #1**: `GH100` is in this enum and in the Providers UI dropdown, but there is no `GH100` entry in `GPU_PRESETS` (in `inference-energy.ts`). If a user selects `GH100`, `getGpuParams()` silently falls back to the H100 default — no error, no warning.

NVIDIA doesn't have a "GH100" product — they have GH200 Grace Hopper. The UI label says "NVIDIA GH100 (144GB)" which matches GH200's specs, so this looks like a naming error.

**Fix**: Either add a `GH100` key to `GPU_PRESETS` or remove it from this enum (and the Providers dropdown). Given the naming confusion, removing `GH100` and keeping `GH200` seems cleaner.

packages/backend/src/services/usage-storage.ts
```diff
@@ -822,4 +844,120 @@ export class UsageStorageService extends EventEmitter {
...
+      // Find all requests for this alias
+      const requests = await db
+        .select({
+          requestId: this.schema.requestUsage.requestId,
+          tokensInput: this.schema.requestUsage.tokensInput,
+          tokensOutput: this.schema.requestUsage.tokensOutput,
+          provider: this.schema.requestUsage.finalAttemptProvider,
+        })
+        .from(this.schema.requestUsage)
+        .where(eq(this.schema.requestUsage.incomingModelAlias, aliasSlug));
```
mcowger: 🔴 **Blocker #2**: This query loads **all** historical requests for an alias into memory with no LIMIT. For busy deployments this could be millions of rows → OOM. The `batchSize = 100` below only controls update concurrency, not the initial SELECT.

**Fix**: Add cursor-based pagination or at minimum iterate with `.limit()` + `.offset()`, e.g.:

```ts
const BATCH_SIZE = 500;
let offset = 0;
while (true) {
  const batch = await db.select({...}).from(...).where(eq(...)).limit(BATCH_SIZE).offset(offset);
  if (batch.length === 0) break;
  // process batch...
  offset += BATCH_SIZE;
}
```

packages/backend/src/routes/management/config.ts
```diff
@@ -153,6 +157,21 @@ export async function registerConfigRoutes(fastify: FastifyInstance) {
     }
     try {
       await configService.saveAlias(slug, result.data);
+
+      // Recalculate energy usage if model_architecture was provided
+      if (result.data.model_architecture && usageStorage) {
+        try {
+          const updated = await usageStorage.recalculateEnergyForAlias(
```
mcowger: 🔴 **Blocker #4**: This recalculation block (lines 161-173) is duplicated verbatim in the PATCH handler below (lines 203-215). Please extract it into a shared helper, e.g.:

```ts
async function recalculateEnergyIfChanged(
  slug: string,
  model_architecture: any,
  usageStorage?: UsageStorageService
) {
  if (model_architecture && usageStorage) {
    try {
      const updated = await usageStorage.recalculateEnergyForAlias(slug, model_architecture);
      logger.info(`Recalculated energy for ${updated} requests for alias '${slug}'`);
    } catch (recalcError) {
      logger.error(`Failed to recalculate energy for alias '${slug}'`, recalcError);
    }
  }
}
```

Then call it from both PUT and PATCH handlers.

packages/backend/src/services/huggingface-model-fetcher.ts
```diff
@@ -0,0 +1,669 @@
...
+ *
+ * Used by the inference energy calculator to get accurate model parameters
+ * (layers, heads, hidden_size, context_length, etc.) for energy estimation.
+ */
+
+import { logger } from '../utils/logger';
+import type { ModelParams } from './inference-energy';
+
+// Common data type sizes in bytes
+export const DTYPE_SIZES: Record<string, number> = {
```
mcowger: 🟡 **Significant #9**: `DTYPE_SIZES` is defined here, but `usage-storage.ts` also has its own `getDtypeSize()` switch statement (line 944) with the same dtype→bytes mapping. If a new dtype is added to one, the other will be out of sync.

**Fix**: Make `inference-energy.ts` the single source of truth — export `DTYPE_SIZES` from there, and import it in both `huggingface-model-fetcher.ts` and `usage-storage.ts`. Remove the `getDtypeSize()` method from `UsageStorageService`.

packages/backend/src/services/inference-energy.ts
```diff
@@ -37,8 +37,120 @@ export interface InferenceFootprint {
...
+  qk_rope_head_dim: 64,
+  dtype_size: 2, // FP16 = 2 bytes
+  heads: 64,
+};
+
+const DEFAULT_GPU: GpuParams = {
+  ram_gb: 192,
+  bandwidth_tb_s: 8.0,
+  flops_tflop: 9000,
+  power_draw_watts: 14300,
```
mcowger: 🟡 **Significant #10**: `DEFAULT_GPU.power_draw_watts: 14300` looks like an 8-GPU H100 cluster (8×700W + overhead), but all the presets define per-GPU power (700W, 1000W, 1400W). The power scaling formula on line 201 does `power_draw_watts * (tp / 8)`, which assumes per-GPU power scaled by TP.

With `DEFAULT_GPU`, this becomes `14300 * (8/8) = 14300W` — accidentally correct for an 8-way cluster, but semantically wrong since the field is called `power_draw_watts` and documented as a single GPU spec.

Meanwhile `getGpuParams()` defaults to a single H100 (700W). These two defaults disagree by ~20×.

**Fix**: Align `DEFAULT_GPU` with the presets — make it a single-GPU spec (e.g., matching B200 at 1000W) or clearly document it as a cluster-level default with matching comment + field name.

packages/backend/src/services/inference-energy.ts
```diff
@@ -37,8 +37,120 @@ export interface InferenceFootprint {
...
+    bandwidth_tb_s: customParams?.bandwidth_tb_s ?? baseGpu.bandwidth_tb_s,
+    flops_tflop: customParams?.flops_tflop ?? baseGpu.flops_tflop,
+    power_draw_watts: customParams?.power_draw_watts ?? baseGpu.power_draw_watts,
+  };
+}
+
+/**
+ * Returns available GPU preset options for UI dropdowns
+ */
+export function getGpuPresetOptions(): Array<{ value: string; label: string }> {
```
mcowger: 🟢 **Minor #12**: `getGpuPresetOptions()` is exported but never called anywhere. The Providers page (`Providers.tsx`) hardcodes its own dropdown options independently. These two lists are already out of sync (this function lacks `GH100`, and the UI has it).

**Fix**: Either wire up this function in the Providers dropdown to keep a single source of truth, or remove it if the UI prefers hardcoded options.

packages/frontend/src/components/EnergyOverTime.tsx
```diff
@@ -0,0 +1,129 @@
...
+    }
+    return timestamp;
+  }
+  return date.toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
+}
+
+/**
+ * Formats kWh values for tooltip display.
+ */
+function formatKwh(value: number): string {
```
mcowger: 🟢 **Minor #14**: `formatKwh()` and `formatTimeLabel()` are local formatters. Per project guidelines (AGENTS.md §8.3), all formatting should be centralized in `packages/frontend/src/lib/format.ts`. The same applies to `formatStreamingTime()` in `EnergyTimeComparison.tsx`.

**Fix**: Add `formatKwh()` and `formatStreamingTime()` (or reuse `formatDuration`) to `format.ts`, then import and use them from both components.

packages/frontend/src/pages/Models.tsx
```diff
@@ -615,6 +657,260 @@ export const Models = () => {
...
+              <div className="flex-1">
+                <label className="font-body text-[11px] font-medium text-text-secondary">
+                  Hugging Face Model ID
+                </label>
+                <input
+                  className="w-full py-2 px-3 font-body text-sm text-text bg-bg-glass border border-border-glass rounded-sm outline-none focus:border-primary"
+                  value={hfModelId}
+                  onChange={(e) => setHfModelId(e.target.value)}
+                  placeholder="e.g. meta-llama/Llama-3.1-70B-Instruct"
+                  onKeyPress={(e) => e.key === 'Enter' && fetchHfModelArchitecture()}
```
mcowger: 🟢 **Nit #15**: `onKeyPress` is deprecated in React. Please use `onKeyDown` instead:

```tsx
onKeyDown={(e) => e.key === 'Enter' && fetchHfModelArchitecture()}
```

Same applies to line 1438 in this file.

packages/backend/src/services/huggingface-model-fetcher.ts
```diff
@@ -0,0 +1,669 @@
...
+  dtype_size: 2, // FP16 = 2 bytes
+  heads: 64,
+};
+
+// ─── Fetcher Implementation ──────────────────────────
+
+export class HuggingFaceModelFetcher {
+  private static instance: HuggingFaceModelFetcher;
+  private cache: Map<string, ModelParams> = new Map();
+  private dtypeCache: Map<string, string> = new Map();
```
mcowger: 🟢 **Minor #16 (no tests)**: There is zero test coverage for the new code in this PR:
- `HuggingFaceModelFetcher` (complex parsing/matching logic — would benefit from unit tests for `parseConfig`, `getHeuristicParams`, `inferDtype`)
- `recalculateEnergyForAlias` in `usage-storage.ts`
- The new `/v0/management/models/huggingface/:modelId` endpoint
- `EnergyOverTime` and `EnergyTimeComparison` frontend components

Please add at least basic unit tests for the HuggingFace fetcher (especially `parseConfig` and `getHeuristicParams`) and the energy recalculation logic.

packages/backend/drizzle/migrations_pg/0029_cloudy_speedball.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE "request_usage" ADD COLUMN "vision_fallthrough_model" text;--> statement-breakpoint
```
mcowger: 🟢 **Minor #17**: This PG migration includes `ALTER TABLE "request_usage" ADD COLUMN "vision_fallthrough_model" text;` which is from a different feature (the vision fallthrough work already merged to main). The SQLite equivalent (migration 0025) was already applied on main.

**Fix**: Please regenerate the PG migration so it only contains the new GPU/model_architecture columns, without the `vision_fallthrough_model` addition. You can do this by:
1. Resetting the PG migration files
2. Ensuring your branch is based on latest main (which already has the `vision_fallthrough_model` column in the schema)
3. Running `bunx drizzle-kit generate --config drizzle.config.pg.ts` to produce a clean migration with only the new columns

mcowger (changes requested):
## Code Review: Energy Usage Improvements

Reviewed the full diff and found several issues that should be addressed before merge. Inline comments below with details.

### Blockers
- #1: `GH100` in config enum and UI dropdown but not in `GPU_PRESETS` — silent fallback to H100
- #2: `recalculateEnergyForAlias` loads all requests into memory with no pagination
- #4: Duplicate recalculation code in PUT and PATCH handlers

### Significant
- #9: `DTYPE_SIZES` duplicated across `huggingface-model-fetcher.ts` and `usage-storage.ts`
- #10: `DEFAULT_GPU` uses cluster-level power (14.3kW) conflicting with per-GPU presets

### Minor
- #12: `getGpuPresetOptions()` exported but never used; already out of sync with UI dropdown
- #14: Energy components use local formatters instead of centralized `format.ts`
- #15 (nit): `onKeyPress` is deprecated, should use `onKeyDown`
- #16: No test coverage for any of the new code
- #17: PG migration bundles unrelated `vision_fallthrough_model` column

alexispurslane:
Goddamn! This is what I get for vibe coding more than usual.

"It'll be quick" I thought. "How bad can it mess up" I thought. 😂

packages/backend/src/services/inference-energy.ts
alexispurslane: Yeah I completely forgot whether I was using cluster or single GPU power in the interem since last touching this math, hence the total inconsistency.

alexispurslane:
@mcowger I think I've addressed all your concerns, and substantially improved the features on offer to boot

mcowger:
Hi @alexispurslane -
Really appreciate the time and thought you put into this PR. I like the idea a lot.

At the moment, I’m trying to avoid adding more visuals to Plexus core. There are already quite a few, and I want to keep the project centered on being an LLM proxy rather than expanding into a visualization engine.

If you want to pursue this direction, I’d recommend using a Prometheus-compatible tool for the visualization layer. And if there’s any data missing that would make that easier, I’d be happy to review a PR for it.

alexispurslane:
If I remove the visualizations component of this PR, would you accept the stuff
that allows the user to set the GPU profile of a provider and the model
architecture information on a model and so on, for Prometheus to use? Or would
you want me to completely rewrite this PR for that tool instead?

Also could I at least keep one visualization card since you already have the
toast stuff in so I could just expand that one card? Or is it less about UX and
more about code volume?

Thanks 


— alexis (she/her)


Sent from Proton Mail for Android.



-------- Original Message --------

mcowger:
> If I remove the visualizations component of this PR, would you accept the stuff that allows the user to set the GPU profile of a provider and the model architecture information on a model and so on, for Prometheus to use? Or would you want me to completely rewrite this PR for that tool instead? Also could I at least keep one visualization card since you already have the toast stuff in so I could just expand that one card? Or is it less about UX and more about code volume? 

Feel free to adjust to just edit the existing card.  

The GPU profile stuff is fine too.

I've adjust some of the ways I'm handling migrations (see CONTRIBUTING.md), which should allow you to make this PR much smaller after rebasing.

alexispurslane:
@mcowger how's this?

alexispurslane:
should probably squash before merge

mcowger:
I'll poke at this tomorrow

mcowger:
Closing this PR as the code has already been integrated under a separate commit. Apologies that this won't show as a merged PR and you won't get the merge credit you deserve — that's on me for not merging this properly at the time. The work itself is very much appreciated though. Thank you @alexispurslane for the great contribution on improving the energy usage calculations and display!

PR #158, opened 2026-04-12
Add key access allowlists and improve logs filtering

zicochaos:
## Summary
- add optional `allowedModels` and `allowedProviders` to API key config and persist them in SQLite/Postgres
- enforce key allowlists during request routing across chat, messages, embeddings, speech, transcriptions, and image routes
- add server-side logs sorting plus an `apiKey` filter in the Logs UI and fix `/ui/*` SPA deep-link handling

## Details
- key allowlists are propagated from auth middleware into request metadata and enforced in the dispatcher before/while selecting candidates
- the Logs page now supports filtering by key alongside model/provider, while retaining server-side sorting for correct pagination
- Keys UI copy was tightened to explain the allowlist behavior more clearly
- added focused backend coverage for key policy propagation, dispatcher enforcement, and usage filtering/sorting

## Migration Note
- migrations were regenerated from schema rather than kept as manual SQL/journal edits
- the generated Postgres migration also includes `request_usage.vision_fallthrough_model`, which Drizzle surfaced as existing schema/migration drift

## Verification
- `docker run --rm -v "$HOME"/Projects/plexus-new:/work -w /work oven/bun:1 bash -lc 'bun install --frozen-lockfile >/tmp/bun-install.log && cd packages/backend && bun test src/routes/inference/__tests__/auth.test.ts src/services/__tests__/dispatcher-failover.test.ts src/services/__tests__/usage-storage-performance.test.ts'`
- `docker run --rm -v "$HOME"/Projects/plexus-new:/work -w /work oven/bun:1 bash -lc 'bun install --frozen-lockfile >/tmp/bun-install.log && cd packages/frontend && bun run build'`
- local Docker verification against a copied live config/SQLite snapshot from `192.168.66.12`:
  - `/ui/logs` loads successfully
  - server-side `apiKey` log filtering works
  - model/provider key restrictions were exercised with blocked and allowed requests

### DIFF
--- a/packages/backend/drizzle/migrations/0026_mean_scrambler.sql
+++ b/packages/backend/drizzle/migrations/0026_mean_scrambler.sql
@@ -1,0 +1,2 @@
+ALTER TABLE `api_keys` ADD `allowed_models` text;--> statement-breakpoint
+ALTER TABLE `api_keys` ADD `allowed_providers` text;
--- a/packages/backend/drizzle/migrations/meta/0026_snapshot.json
+++ b/packages/backend/drizzle/migrations/meta/0026_snapshot.json
@@ -1,0 +1,2422 @@
+{
+  "version": "6",
+  "dialect": "sqlite",
+  "id": "6f2bbe97-9cf7-4542-8eb5-f03a3fc0b56e",
+  "prevId": "7bd18f77-7911-41c7-b925-8c5f20356808",
+  "tables": {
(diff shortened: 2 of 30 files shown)


mcowger:
Really nicely done PR, and thanks for the SPA fix - I had been meaning to get to that.  

I'm going to follow it up with a UI change to make the selectors drop downs with filtering, and then will cut a release.

PR #160, opened 2026-04-14
feat: allow API keys to log into admin panel with limited scope

darkspadez:
The following PR allows for users to login with their api key and view their logs, statistics, and enable/disable trace. Also allows them to rotate their key.


mcowger:
Thanks for this PR — the overall approach is solid and the self-service UX for limited users is well thought out. I've rebased the branch onto main and regenerated the migrations properly (SQLite `0027_green_emma_frost`, PG `0031_last_wallow`) to resolve the numbering collision with `0030_acoustic_katie_power`. No logic changes — just clean drizzle-kit output.

Two things need to be addressed before this can merge:

---

### 1. Events SSE endpoint leaks all keys' activity to limited users

**File:** `packages/backend/src/routes/management/usage.ts` — the `GET /v0/management/events` endpoint

The endpoint sits inside the `authenticate`-scoped plugin, so it correctly rejects unauthenticated requests. However, it attaches global listeners on `usageStorage` (`started`, `completed`, `error`) and forwards every event to the SSE client without filtering. A limited user who subscribes to this stream sees real-time request activity — model, provider, token counts, latency — for **all** API keys in the system, not just their own.

This breaks the isolation guarantee that is the whole point of limited-principal scoping. An adversarial key holder could passively monitor the usage patterns, provider availability, and error rates of every other key.

**Fix:** Call `scopedKeyName(request)` (already exported from `_principal.ts`) when the handler sets up its listeners, and skip emitting any event whose `apiKey` field doesn't match the scoped key. Admin principals (where `scopedKeyName` returns `null`) should continue to receive all events unchanged.

---

### 2. Cooldown clear is accessible to all limited users regardless of `allowedProviders`

**File:** `packages/backend/src/routes/management/cooldowns.ts` — the `DELETE /v0/management/cooldowns/:provider` endpoint

The current guard is:
```ts
if (allowed.length > 0 && !allowed.includes(provider)) {
  return reply.code(403)...
}
```

When `allowedProviders` is an empty array (meaning "no provider restriction"), `allowed.length > 0` is false and the check is skipped entirely — so the limited user can clear the cooldown for any provider. But clearing a provider cooldown is a system-level action with real blast radius: it forces the router to retry a provider that may be cooling down for good reason (rate limit, outage, quota exhaustion). This is not a safe self-service operation for key holders.

**Fix:** Make both cooldown-clear endpoints (`DELETE /cooldowns/:provider` and `DELETE /cooldowns`) admin-only by adding `requireAdmin` to those specific routes, or by moving them into the `adminOnly` plugin in `management.ts`. The limited-user scoping logic in the handler body can then be removed entirely.

---

Everything else looks good. Happy to re-review once these two are addressed.

darkspadez:
@mcowger all ready for ya!

mcowger:
@claude review this please, with a valley girl style.

PR #163, opened 2026-04-14
feat: add metadata overrides and custom source for model aliases

darkspadez:
Lets users override individual fields (name, description, context length,
pricing, architecture, supported parameters, top provider limits) on top of
models.dev / OpenRouter / Catwalk catalog data, and adds a new 'custom'
metadata source where users supply everything themselves — seeded with
sensible defaults.

### DIFF
--- a/packages/backend/src/db/config-repository.ts
+++ b/packages/backend/src/db/config-repository.ts
@@ -18,6 +18,7 @@ import type {
   McpServerConfig,
   FailoverPolicy,
   CooldownPolicy,
+  MetadataOverrides,
 } from '../config';
 
 // Helper to parse JSON from SQLite text columns (PG jsonb auto-deserializes)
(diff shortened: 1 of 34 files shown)


packages/backend/src/db/config-repository.ts
```diff
@@ -429,7 +480,8 @@ export class ConfigRepository {
         .where(eq(schema.modelAliasTargets.aliasId, row.id))
         .orderBy(schema.modelAliasTargets.sortOrder);
 
-      result[row.slug] = this.rowToModelConfig(row, targets);
+      const overrideRow = await this.getMetadataOverrideRow(row.id);
+      result[row.slug] = this.rowToModelConfig(row, targets, overrideRow);
```
mcowger, line 507: This adds a second sequential query per alias inside the existing loop, turning the already-N+1 targets fetch into 1+2N round-trips total. Please batch-fetch all override rows before the loop (e.g. with `inArray` across all alias IDs, run in parallel with the targets fetch) and then look them up by alias ID in a Map.

packages/frontend/src/pages/Debug.tsx
```diff
@@ -264,81 +266,85 @@ export const Debug: React.FC = () => {
         <div>
           <h1 className="font-heading text-3xl font-bold text-text m-0 mb-2">Debug Traces</h1>
           <p className="text-[15px] text-text-secondary m-0">
-            Inspect full request/response lifecycles
+            {isLimited && principal?.keyName
+              ? `Traces for key "${principal.keyName}" only. Toggle capture in My Key.`
```
mcowger, line 270: TypeScript cannot narrow the `Principal` union through the opaque `isLimited` boolean, so `principal?.keyName` is a type error here (`keyName` only exists on the `'limited'` variant). Same issue in Errors.tsx and Logs.tsx. Use the discriminant directly: `principal?.role === 'limited' && principal.keyName`.

packages/backend/src/db/config-repository.ts
mcowger, line 507: This adds a second sequential query per alias inside the existing loop, turning the already-N+1 targets fetch into 1+2N round-trips total. Please batch-fetch all override rows before the loop (e.g. with `inArray` across all alias IDs, run in parallel with the targets fetch) and look them up by alias ID in a Map.

packages/frontend/src/pages/Debug.tsx
mcowger, line 270: TypeScript cannot narrow the `Principal` union through the opaque `isLimited` boolean, so `principal?.keyName` is a type error here (`keyName` only exists on the `'limited'` variant). Same issue in Errors.tsx and Logs.tsx. Use the discriminant directly: `principal?.role === 'limited' && principal.keyName`.

packages/backend/src/db/config-repository.ts
```diff
@@ -429,7 +480,8 @@ export class ConfigRepository {
         .where(eq(schema.modelAliasTargets.aliasId, row.id))
         .orderBy(schema.modelAliasTargets.sortOrder);
 
-      result[row.slug] = this.rowToModelConfig(row, targets);
+      const overrideRow = await this.getMetadataOverrideRow(row.id);
```
mcowger: This adds a second sequential query per alias inside the existing loop, turning the already-N+1 targets fetch into 1+2N round-trips total. Please batch-fetch all override rows before the loop (e.g. with `inArray` across all alias IDs, run in parallel with the targets fetch) and look them up by alias ID in a Map.

packages/frontend/src/pages/Debug.tsx
```diff
@@ -264,81 +266,85 @@ export const Debug: React.FC = () => {
         <div>
           <h1 className="font-heading text-3xl font-bold text-text m-0 mb-2">Debug Traces</h1>
           <p className="text-[15px] text-text-secondary m-0">
-            Inspect full request/response lifecycles
+            {isLimited && principal?.keyName
```
mcowger: TypeScript cannot narrow the `Principal` union through the opaque `isLimited` boolean, so `principal?.keyName` is a type error here (`keyName` only exists on the `'limited'` variant). Same issue in Errors.tsx and Logs.tsx. Use the discriminant directly: `principal?.role === 'limited' && principal.keyName`.

packages/backend/src/db/config-repository.ts
mcowger: This adds a second sequential query per alias inside the existing loop, turning the already-N+1 targets fetch into 1+2N round-trips total. Please batch-fetch all override rows before the loop (e.g. with `inArray` across all alias IDs, run in parallel with the targets fetch) and look them up by alias ID in a Map.

mcowger (commented):
Two issues introduced by this PR:

**1. N+1 query worsened in `getAllAliases()` (`config-repository.ts:483`)**

The PR adds a `getMetadataOverrideRow(row.id)` call inside the existing per-alias loop, which already fires one targets query per alias. This turns 1+N into 1+2N sequential round-trips. Please batch-fetch all override rows before the loop using `inArray` across all alias IDs (can be run in parallel with the targets fetch), then look them up by alias ID in a Map.

**2. TypeScript type errors on `principal?.keyName` (`Debug.tsx:269`, `Errors.tsx:149`, `Logs.tsx:391`)**

`keyName` only exists on the `'limited'` variant of the `Principal` union — not on `{ role: 'admin' }`. TypeScript cannot narrow the union through the opaque `isLimited` boolean, so these three lines are type errors. Use the discriminant directly instead: `principal?.role === 'limited' && principal.keyName`.

mcowger:
@claude review this please, like a pirate.

PR #167, opened 2026-04-15
feat: add per-alias enforce_limits toggle for pre-dispatch context check

darkspadez:
Adds an opt-in boolean on model aliases that runs a fast token-estimation
check before dispatching to the upstream provider. When enabled, the
dispatcher rejects locally with a 400 context_length_exceeded error if the
estimated input tokens plus reserved output tokens exceed the model's
context window — avoiding a wasted upstream round-trip and an opaque
provider-side 400.

Behavior:
- Toggle lives on ModelConfig alongside use_image_fallthrough (not in
  metadata.overrides) — it is routing policy, not catalog data.
- Reservation uses min(request.max_tokens, metadata.max_completion_tokens)
  to minimize false rejections when the caller asked for a small
  completion.
- Fails open (with a debug log) when no context_length is known — can't
  enforce what we don't know.
- Reuses existing estimateInputTokens() heuristic (microseconds, no WASM)
  with a 10% safety multiplier to cover the estimator's ±20–30% variance.
- The context_length_exceeded code propagates through each endpoint's
  native error envelope (chat/messages/responses/gemini).

Includes 12 new tests covering toggle on/off, oversized/under-limit,
missing metadata (fail-open), max_tokens-vs-metadata precedence, and all
four API shapes (chat, messages, gemini, responses).

### DIFF
--- a/packages/backend/drizzle/schema/postgres/model-aliases.ts
+++ b/packages/backend/drizzle/schema/postgres/model-aliases.ts
@@ -31,6 +31,7 @@ export const modelAliases = pgTable('model_aliases', {
   useImageFallthrough: boolean('use_image_fallthrough').notNull().default(false),
   // Model architecture override for inference energy calculation
   modelArchitecture: jsonb('model_architecture'), // override for total_params, active_params, layers, heads, kv_lora_rank, qk_rope_head_dim, context_length, dtype
+  enforceLimits: boolean('enforce_limits').notNull().default(false),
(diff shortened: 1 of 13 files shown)


mcowger:
Please rebase this on main.  

ALso, I've changed how I'm doing migrations - you'll no longer commit the journals, sql files etc.  Please see CONTRIBUTING.md.

Also, please run a `bun install` after your rebase, which will configure the required git hooks and stuff

PR #184, opened 2026-04-16
Add Neuralwatt balance and subscription quota checkers

### LINKED ISSUE #169: Add Neuralwatt balance checker
I think this might need to use the `X-Allowance-Remaining-USD` header included in completions responses. I don't see some dedicated way to just "check balance" without a completions request, but maybe I'm missing it.

https://portal.neuralwatt.com/docs/guides/allowances

mcowger:
Implements two quota checkers for the Neuralwatt API (/v1/quota endpoint):

- `neuralwatt`: balance checker tracking credit balance (credits_remaining_usd)
- `neuralwatt-subscription`: rate-limit checker tracking kWh subscription usage

Both use Bearer token auth against api.neuralwatt.com/v1/quota.
Includes frontend display components, config components, and integration
with all relevant frontend pages and cards.

Closes #169

Generated with [Claude Code](https://claude.ai/code)


mcowger:
@claude I think we need to modify this to a single checker that populates both types so that a user doesn't have to double configure just to get both checker types.

PR #205, opened 2026-04-17
Collect kwh data from neuralwatt response

### LINKED ISSUE #204: Collect kwh data from neuralwatt response
We currently collect cost data from streaming responses like this (where the cost sometimes come in as a SSE comment):

: cost {"request_cost_usd": 0.0002645210159619417, "cache_savings_usd": 0.0, "allowance_remaining_usd": 14.8763, "budget_remaining_usd": 14.8763}

Similarly, we also have the following energy data:

: energy {"energy_joules": 190.46, "energy_kwh": 5.2904e-05, "avg_power_watts": 3109.0, "duration_seconds": 0.613, "attribution_method": "prorated_multi_gpu_8", "attribution_ratio": 0.1, "ratio_was_capped": true, "uncapped_attribution_ratio": 1.0, "uncapped_energy_joules": 1904.55, "uncapped_energy_kwh": 0.000529042}

We should extract the 'energy_kwh' (converting from scientific notation) and store that as the energy usage where available instead of the estimate we currently store.

mcowger:
Extract energy_kwh from SSE energy comments and store as the energy usage where available instead of the estimate we currently store.

Closes #204

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/backend/src/services/inspectors/debug-logging.ts
+++ b/packages/backend/src/services/inspectors/debug-logging.ts
@@ -147,6 +147,30 @@ export class DebugLoggingInspector extends BaseInspector {
     return lastCost;
   }
 
+  /**
+   * Extract provider-reported energy from SSE comment lines (e.g., neuralwatt).
+   * Providers emit `: energy {"energy_kwh": ...}` as SSE comments
+   * alongside the standard `data:` events. These contain actual energy usage.
+   */
+  private extractProviderEnergyFromSSEComments(fullBody: string): any | null {
+    const lines = fullBody.split(/\r?\n/);
+    let lastEnergy: any = null;
+
+    for (const line of lines) {
+      // Match `: energy {json}` pattern (SSE comment with energy data)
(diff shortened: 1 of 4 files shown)


packages/backend/src/services/inspectors/debug-logging.ts
```diff
@@ -147,6 +147,30 @@ export class DebugLoggingInspector extends BaseInspector {
...
+  /**
+   * Extract provider-reported energy from SSE comment lines (e.g., neuralwatt).
+   * Providers emit `: energy {"energy_kwh": ...}` as SSE comments
+   * alongside the standard `data:` events. These contain actual energy usage.
+   */
+  private extractProviderEnergyFromSSEComments(fullBody: string): any | null {
+    const lines = fullBody.split(/\r?\n/);
+    let lastEnergy: any = null;
+
+    for (const line of lines) {
```
mcowger, line 159: @claude wouldn't it be better to strip the prefix ": energy " and then parse the string as JSON?

PR #214, opened 2026-04-19
ci(pi-agent): fix silent failure via local in-place baseUrl override

mcowgeratsigma:
## Summary

- Replaces the remote `pi-env-var-provider` extension with a local one that mutates the pi-ai built-in openrouter model objects **in place** instead of relying on `pi.registerProvider()`'s override mode.
- Fixes the silent failure where `@pi` ran successfully but never posted any comment.

## Root cause

`pi-coding-agent-action@v2.8.1`'s `Agent` constructor calls `modelRegistry.find(provider, model)` and caches the returned `Model` reference to `this.model` **before any extension runs**. When an extension later calls `pi.registerProvider("openrouter", { baseUrl })`, `applyProviderConfig` in override mode replaces each model via `this.models = this.models.map(m => ({ ...m, baseUrl: ... }))`. That produces brand-new objects, so `Agent.this.model` still points at the original object with `baseUrl = "https://openrouter.ai/api/v1"`. Requests go to openrouter.ai with the Plexus API key and are rejected — silently, since no output chunks arrive.

## Fix

The new `.github/pi-extensions/plexus-override` extension imports `getModels` from `@mariozechner/pi-ai`, which returns references into the singleton `modelRegistry` Map. Mutating `model.baseUrl` in place updates every reference — including the Agent's cached `this.model` — so the request hits the Plexus proxy with the Plexus key.

## Test plan

- [ ] Post a comment like `@pi say hi` on any PR/issue and verify the action replies.
- [ ] Confirm the Plexus proxy logs show the request (not openrouter.ai).

https://claude.ai/code/session_01684L3rjEZTNuceLL99odwU

### DIFF
--- a/.github/pi-extensions/plexus-override/index.js
+++ b/.github/pi-extensions/plexus-override/index.js
@@ -1,0 +1,46 @@
+// Plexus proxy baseUrl override for the Pi coding agent.
+//
+// Problem: pi-coding-agent-action's Agent constructor calls
+// modelRegistry.find() and caches the returned Model reference BEFORE any
+// extension factory runs. If we then call pi.registerProvider() in
+// override mode, the registry replaces its model objects via `.map(...)`
+// with new objects, leaving the Agent's captured reference stale. The
+// provider (openai-completions) reads baseUrl from the captured model at
+// request time and sends the request to the original openrouter.ai URL
+// with the Plexus API key, which fails silently.
+//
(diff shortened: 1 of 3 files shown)


mcowgeratsigma:
@pi please reply with a short greeting so we can verify the Plexus proxy is now being used. If you see this, say "hello from Plexus" and nothing else.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #276, opened 2026-04-25
feat(quota): implement Meter model redesign (QUOTA_BALANCE_REDESIGN)

mcowgeratsigma:
Replaces the old QuotaWindow[]-based system with a flat Meter[] model
where each meter is self-describing with kind, unit, period fields, and
status — eliminating per-checker UI switches, checker-level categories,
compound enum window types, and 21 near-duplicate display components.

Backend changes:
- New Meter/MeterCheckResult types in src/types/meter.ts
- Self-registering checker-registry.ts (defineChecker + loadAllCheckers)
- New meter_snapshots table schema (SQLite + PostgreSQL)
- All 22 checkers rewritten to export default defineChecker({...})
- apertis and apertis-coding-plan merged into single apertis checker
- quota-scheduler.ts rewired to use registry + persist to meter_snapshots
- /v0/management/quotas routes return MeterCheckResult shape
- apertis-coding-plan removed from config schema

Frontend changes:
- New QuotaCheckerInfo type with meters: Meter[] replacing latest: snapshot[]
- New generic components: BalanceMeterRow, AllowanceMeterRow, MeterValue
- checker-presentation.ts for display names
- Quotas.tsx, CombinedBalancesCard, CompactQuotasCard, CompactBalancesCard
  rewritten to use meters array directly
- Sidebar.tsx updated to derive balance/allowance from meter kinds
- All 21 XxxQuotaDisplay.tsx components deleted
- QuotaHistoryModal, BalanceHistoryModal deleted
- api.ts updated to use new response shape

https://claude.ai/code/session_01LUsqANXxpFLZA61deDSnGF


mcowger:
**Claude finished @mcowger's task in 13m 42s** —— [View job](https://github.com/mcowger/plexus/actions/runs/24966821238)

---
### PR Review: feat(quota): implement Meter model redesign

Reviewing 92 files changed (+3,395 / −8,884). The core architecture is clean and well-structured — the Meter model is a significant improvement over the old QuotaWindow system. Below are my findings.

---

### Critical Issues

**1. Missing database migration for `meter_snapshots` table**

The `meter_snapshots` table schema is defined in Drizzle (`schema/sqlite/meter-snapshots.ts`, `schema/postgres/meter-snapshots.ts`) but there is **no migration** that actually creates this table. The latest migration snapshot (`0032_snapshot.json`) does not include `meter_snapshots`. This means the table won't exist in production unless `runMigrations()` generates a push-based migration at runtime. If you're using Drizzle Kit push mode, this may work in dev, but it will break on any deployed instance using standard migrations.

The `quota-scheduler.ts:persistResult()` writes to `schema.meterSnapshots` — this will fail at runtime with a "table not found" error on any database that hasn't been set up with push mode.

[Fix this →](https://claude.ai/code?q=Generate%20a%20Drizzle%20migration%20for%20the%20new%20meter_snapshots%20table%20by%20running%20drizzle-kit%20generate%20in%20packages/backend%20so%20the%20table%20is%20created%20on%20deployed%20instances.&repo=mcowger/plexus)

**2. `exhaustionThreshold` field on `Meter` type but not persisted to DB**

The `Meter` interface (`meter.ts:38`) includes `exhaustionThreshold`, the registry passes it through (`checker-registry.ts:142,166`), but neither the SQLite nor PostgreSQL `meter_snapshots` table schema has an `exhaustion_threshold` column. The `quota-scheduler.ts:persistResult()` method does not persist this field. This means:
- The value is set at check time but lost on read-back from `getLatestQuota()`
- Any downstream consumer relying on per-meter exhaustion thresholds (instead of the per-config `maxUtilizationPercent`) will get `undefined`

Either add the column to both schemas and the persist logic, or remove `exhaustionThreshold` from `Meter` since the scheduler already uses `config.options.maxUtilizationPercent` for cooldown gating.

[Fix this →](https://claude.ai/code?q=Either%20add%20exhaustion_threshold%20column%20to%20both%20meter-snapshots%20table%20schemas%20(sqlite%20and%20postgres)%20and%20persistResult()%20in%20quota-scheduler.ts%2C%20or%20remove%20the%20exhaustionThreshold%20field%20from%20the%20Meter%20interface%20in%20meter.ts%20since%20cooldown%20logic%20already%20uses%20config.options.maxUtilizationPercent.&repo=mcowger/plexus)

---

### Important Issues

**3. `deriveUtilization` double-counts `used` when both `used/limit` and `used+remaining` are available**

In `checker-registry.ts:77-90`, when `limit > 0` and `used` are both defined, the function uses `used/limit` and **ignores** `remaining`. But the `Meter` type and `ctx.allowance()` often provide all three (`used`, `limit`, `remaining`). If the remote API's `remaining` doesn't perfectly equal `limit - used` (due to rounding, async lag, etc.), the utilization percentage computed from `used/limit` will disagree with the `remaining` value stored on the meter. This could cause confusing UI display (e.g., 80% utilization shown alongside a `remaining` value that implies 70% utilization).

Consider falling back to `remaining / (used + remaining)` only when `limit` is absent, or documenting that `limit` and `used` are the source of truth.

**4. `getLatestQuota` fetches 200 rows and filters in-memory by exact timestamp match**

`quota-scheduler.ts:284-295` fetches the 200 most recent rows and groups by `checkedAt`. This works for small installations but has two issues:
- The `toMs(rows[0].checkedAt)` exact-equality filter on line 295 could miss meters from the same check if the database loses sub-millisecond precision (e.g., SQLite's `timestamp_ms` mode rounding). This would cause `getLatestQuota()` to return a partial meter set.
- On installations with many meters per check, the 200-row limit could be too low if checks happen frequently.

Consider querying with a `GROUP BY checker_id, checked_at` approach or using a dedicated "latest check" query.

**5. `Quotas.tsx` shows `toTitleCase(quota.checkerId)` instead of `getCheckerDisplayName`**

In `Quotas.tsx`, the `CombinedBalancesCard` display uses `toTitleCase(quota.checkerId)` (via `CombinedBalancesCard.tsx:40`) while `CompactQuotasCard` uses `getCheckerDisplayName(quota.checkerType, quota.checkerId)`. This inconsistency means the balance section shows raw IDs like `my_provider` title-cased, while the allowance section shows the proper display name like "OpenRouter".

[Fix this →](https://claude.ai/code?q=In%20CombinedBalancesCard.tsx%20line%2040%2C%20replace%20toTitleCase(quota.checkerId)%20with%20getCheckerDisplayName(quota.checkerType%2C%20quota.checkerId)%20for%20consistency%20with%20CompactQuotasCard.tsx.%20Import%20getCheckerDisplayName%20from%20./checker-presentation.&repo=mcowger/plexus)

**6. `CompactBalancesCard` shows remaining but not `used/limit` when `remaining` is undefined**

`CompactBalancesCard.tsx:33` uses `balanceMeter?.remaining` as the sole display value. For balance meters where `remaining` is `undefined` but `used` and `limit` are both present (a valid Meter state), the card shows "—" instead of computing `limit - used`. `BalanceMeterRow` correctly handles this fallback (`line 14`), but the compact version does not.

[Fix this →](https://claude.ai/code?q=In%20CompactBalancesCard.tsx%2C%20add%20fallback%20for%20remaining%20when%20undefined%3A%20add%20after%20line%2032%3A%20const%20used%20%3D%20balanceMeter%3F.used%3B%20const%20limit%20%3D%20balanceMeter%3F.limit%3B%20const%20displayValue%20%3D%20remaining%20!%3D%3D%20undefined%20%3F%20remaining%20%3A%20(used%20!%3D%3D%20undefined%20%26%26%20limit%20!%3D%3D%20undefined%20%3F%20limit%20-%20used%20%3A%20undefined)%3B%20and%20update%20formattedBalance%20to%20use%20displayValue%20instead%20of%20remaining.&repo=mcowger/plexus)

---

### Minor Issues / Suggestions

**7. Old `quota_snapshots` table not cleaned up**

The `quota_snapshots` table schema files still exist (`sqlite/quota-snapshots.ts`, `postgres/quota-snapshots.ts`) and are still exported from the schema indexes. There's no migration to drop the old table. This isn't blocking but leaves orphaned tables/columns in production databases. Consider a cleanup migration post-merge.

**8. `loadAllCheckers()` import list must be kept in sync manually**

`checker-registry.ts:176-199` uses explicit dynamic imports for each checker. Adding a new checker requires updating this list — forgetting to do so means the checker silently won't load. Consider using `import.meta.glob` (Vite/Node) or a directory scan approach to auto-discover checkers.

**9. No input validation on `/v0/management/quotas/:checkerId/history` since parameter**

`quotas.ts:86-90` parses the `since` query parameter with `parseFloat` for the `d` suffix case, but doesn't validate that the result is a positive number or that the date string is parseable. Passing `since=abc` would produce `NaN` from `new Date('abc').getTime()`, which would then be passed to `toDbTimestampMs`. This is a minor robustness concern.

**10. Frontend `Meter` type diverges from backend**

`frontend/src/types/quota.ts:Meter` lacks `exhaustionThreshold` that the backend `Meter` type has. This is actually fine since the frontend currently doesn't use it, but it means the types are out of sync. If `exhaustionThreshold` is kept on the backend (see issue #2), consider adding it to the frontend type.

**11. Duplicate MeterKind/Utilization/MeterStatus types between backend and frontend**

`backend/src/types/meter.ts` and `frontend/src/types/quota.ts` both define `MeterKind`, `MeterStatus`, `Utilization`, and `Meter` independently. These could drift over time. Consider extracting to the `@plexus/shared` package.

---

### Test Coverage

The test coverage is solid — all 22 checkers have updated test files, and the `quota-scheduler.test.ts` covers the key persistence and cooldown scenarios. A few gaps:

- No test for `getLatestQuota()` — the complex timestamp-matching logic at `quota-scheduler.ts:275-348` is untested. This is the most fragile part of the scheduler.
- No test for `getQuotaHistory()`.
- No test for the management routes (`quotas.ts`).

---

### Architecture Summary

The redesign is well-executed. The `Meter` model (balance/allowance with self-describing kind, unit, period) is much simpler than the old `QuotaWindow[]` system. The `defineChecker()` + registry pattern is clean and consistent across all 22 checkers. The elimination of 21 per-checker `*QuotaDisplay.tsx` components in favor of `BalanceMeterRow` + `AllowanceMeterRow` is a major win.

The critical blocker is the missing migration — without it, the `meter_snapshots` table won't exist on deployed instances.

mcowger:
**Claude finished @mcowger's task in 4m 37s** —— [View job](https://github.com/mcowger/plexus/actions/runs/24966944128)

---
### PR Review: feat(quota): implement Meter model redesign

Reviewing 92 files changed (+3,395 / −8,884). Deep analysis complete.

---

#### Overall Assessment

This is a well-executed architectural simplification. The Meter model replaces a fragmented, per-checker UI system with a self-describing, generic data model. The code reduction is dramatic (net −5,489 lines) and the new design is significantly more maintainable. The core types, registry pattern, and generic frontend components are solid.

That said, I found several substantive issues that should be addressed before merging.

---

#### Bugs & Logic Errors

**1. `balance()` never computes utilization from `remaining` alone**
`checker-registry.ts:122-126` — When a balance meter provides only `remaining` (no `limit`/`used`), `deriveUtilization` returns `'not_applicable'`. But the `balance()` path short-circuits: if `limit` AND `used` are both present it calls `deriveUtilization`, otherwise it hardcodes `'not_applicable'`. This means a balance with only `remaining` and `used` (but no `limit`) gets `'not_applicable'` instead of computing utilization from `used + remaining`, even though `deriveUtilization` supports that case. The `allowance()` path handles this correctly by always calling `deriveUtilization`.

[Fix this →](https://claude.ai/code?q=In%20packages%2Fbackend%2Fsrc%2Fservices%2Fquota%2Fchecker-registry.ts%2C%20the%20balance%28%29%20method%20around%20line%20122%20should%20call%20deriveUtilization%28params.used%2C%20params.limit%2C%20params.remaining%29%20unconditionally%20instead%20of%20short-circuiting%20to%20%27not_applicable%27%20when%20both%20limit%20and%20used%20are%20missing.%20This%20matches%20how%20allowance%28%29%20works%20on%20line%20147.&repo=mcowger/plexus)

**2. `getLatestQuota` filters rows by exact `checkedAt` timestamp — fragile for PostgreSQL**
`quota-scheduler.ts:294-295` — Rows are fetched then filtered with `toMs(r.checkedAt) === latestMs`. For PostgreSQL where `checkedAt` is a `bigint` (epoch ms), `toMs` parses it through `new Date(val).getTime()`, which works. But the equality comparison between two `Date` objects created from the same bigint could theoretically diverge due to floating point in edge cases. More practically, if two check runs land at the same millisecond but one write is slower, the `LIMIT 200` + filter could split a result set. Consider also matching on a `batchId` or grouping by `(checkerId, checkedAt)` in the query itself.

**3. `Meter.exhaustionThreshold` is defined but never read by anyone**
`meter.ts:38` defines `exhaustionThreshold` on the `Meter` interface, and it's passed through in `checker-registry.ts:142,166` — but `quota-scheduler.ts:139` reads the threshold exclusively from `config.options.maxUtilizationPercent`, never from `meter.exhaustionThreshold`. This field is dead code that will mislead future readers into thinking meters can override the threshold per-meter. Either wire it up or remove it.

[Fix this →](https://claude.ai/code?q=Either%20remove%20the%20exhaustionThreshold%20field%20from%20the%20Meter%20interface%20in%20packages%2Fbackend%2Fsrc%2Ftypes%2Fmeter.ts%20line%2038%20and%20from%20the%20BalanceParams%2FAllowanceParams%20in%20checker-registry.ts%2C%20or%20wire%20it%20up%20in%20quota-scheduler.ts%20applyCooldownsFromResult%20so%20each%20meter%20can%20override%20the%20config-level%20threshold.%20Currently%20it%27s%20dead%20code.&repo=mcowger/plexus)

---

#### Security & Data Integrity

**4. No input validation on `checkerId` route parameter**
`quotas.ts:64,78,108` — The `checkerId` from `request.params` is used directly in DB queries and log messages without any validation. While the DB layer uses parameterized queries (safe from injection), an unchecked `checkerId` could contain path-traversal-like strings that pollute logs or be used for reconnaissance. Consider validating it matches a safe pattern (e.g., `^[a-z0-9_-]+$`).

**5. `getOption`/`requireOption` use unsafe type assertions**
`checker-registry.ts:110,115` — `options[key] as T` is an unchecked type assertion. If a config file provides a string where a number is expected, this won't throw — it'll silently propagate the wrong type downstream. The `optionsSchema` on each checker definition validates at... no point, actually. The schema is defined but never used to validate the options before they're passed to `check()`. This is a runtime safety gap.

[Fix this →](https://claude.ai/code?q=In%20packages%2Fbackend%2Fsrc%2Fservices%2Fquota%2Fquota-scheduler.ts%2C%20inside%20runCheckNow%28%29%20around%20line%2099%2C%20parse%20config.options%20through%20def.optionsSchema.parse%28%29%20before%20passing%20to%20createMeterContext%28%29.%20This%20ensures%20each%20checker%27s%20zod%20schema%20is%20actually%20enforced%20at%20runtime.&repo=mcowger/plexus)

---

#### Design Concerns

**6. Self-registering checkers via module-level side effects**
`checker-registry.ts:176-199` — Each checker calls `defineChecker()` at module load time, which mutates a module-scoped `Map`. The `loadAllCheckers()` function imports every checker explicitly. This pattern:
- Makes it easy to forget adding a new checker to `loadAllCheckers()`
- Creates invisible coupling between the registry and checker files
- Makes testing harder (checkers registered in one test leak to others)

Consider a more explicit registration approach (e.g., returning an array from `loadAllCheckers`) or at minimum adding a test that `getCheckerTypes()` length matches the number of checker files.

**7. Frontend `Meter` type duplicates backend instead of sharing**
`packages/frontend/src/types/quota.ts` re-declares `Meter`, `MeterKind`, `MeterStatus`, `Utilization` — duplicating `packages/backend/src/types/meter.ts`. The frontend also omits `PeriodUnit`, `PeriodCycle` types and the `exhaustionThreshold` field. If these diverge further, it'll cause subtle bugs. Consider extracting shared types to `@plexus/shared` or at least documenting the contract.

**8. `primaryAllowanceMeter` picks "most constrained" by raw utilization — can be misleading**
`CompactQuotasCard.tsx:13-19` — The function picks the meter with the highest `utilizationPercent`, but for checkers like `antigravity` that emit dozens of per-model meters (some near 100% by design), this will always surface the worst single model rather than a representative "are we OK?" signal. Consider either using the first meter or filtering to meters with `scope === undefined` (i.e., top-level rather than per-model).

---

#### Missing Test Coverage

**9. No tests for `createMeterContext().balance()` with partial data**
The test files cover individual checkers well, but there's no unit test for the context factory itself — specifically:
- `balance()` with only `remaining` (no `limit`/`used`) — should get `not_applicable`
- `balance()` with `used` + `remaining` but no `limit` — should compute utilization
- `deriveStatus()` at exact boundary values (75, 90, 100)
- `getOption` with missing key returning default
- `requireOption` throwing on missing key

**10. No test for `loadAllCheckers()` completeness**
There should be a test verifying that every checker file in `checkers/` is imported by `loadAllCheckers()`, to prevent future checkers from being silently skipped.

---

#### Minor Issues

**11. `any` types in `getLatestQuota` and `getQuotaHistory`**
`quota-scheduler.ts:291,295,310,377` — Several `as any[]` and `as any` casts. These bypass type safety for DB row access. Consider defining a typed row type or using Drizzle's inferred types.

**12. Stale AGENTS.md reference to `ApertisCodingPlanQuotaDisplay`**
`packages/frontend/src/components/quota/AGENTS.md:73` still references the deleted `ApertisCodingPlanQuotaDisplay` pattern. This should be updated to reflect the new Meter-based approach.

**13. `quota_snapshots` table still exists in both schemas**
The old `quota_snapshots` table schema files still exist and are still exported in both `sqlite/index.ts` and `postgres/index.ts`. If this table is truly superseded, consider adding a migration to drop it and removing the schema exports. If it's retained for backward compatibility, document why.

**14. 15-second hardcoded DB timeout**
`quota-scheduler.ts:280,366` — The `setTimeout(() => reject(...), 15000)` pattern for DB queries creates unhandled promise rejections if the query completes after timeout (the DB connection is still active). Consider using `AbortController` or drizzle's built-in timeout support.

---

#### What Looks Good

- The `defineChecker` + `createMeterContext` pattern eliminates massive duplication
- The `Meter` type with `kind`/`unit`/`period*` is self-describing and extensible
- Generic `AllowanceMeterRow` / `BalanceMeterRow` / `MeterValue` replace 21 near-duplicate display components
- The cooldown logic correctly handles the strictest-wins and lenient-can't-clear-strict cases
- DB schema indices are well-chosen for the query patterns
- Test coverage for the scheduler's cooldown/threshold logic is thorough
- The `apertis` + `apertis-coding-plan` merge is clean

---

mcowger:
Committed in 300

PR #296, opened 2026-04-26
feat: auto-create release tag if it doesn't exist

### LINKED ISSUE #295: Release.yml requires pre-existing tag
The current release.yml requires that the tag is pre-existing when run manually.

It should be updated so that if the tag doesn't exist already, it will tag the current HEAD of main as the given tag, and continue with the release process.

mcowger:
## Summary

- When triggering `release.yml` via `workflow_dispatch`, the tag no longer needs to exist beforehand
- `release-setup.yml` now checks if the tag exists and, if missing, creates it on the current HEAD (main) and pushes it to origin
- Updated the `workflow_dispatch` input description to reflect the new behavior
- Added `contents: write` permission to the setup job so it can push the new tag

Closes #295

---

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.github/workflows/release-setup.yml
+++ b/.github/workflows/release-setup.yml
@@ -15,6 +15,8 @@ jobs:
   setup:
     name: Setup
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     outputs:
       tag: ${{ steps.release_tag.outputs.tag }}
     steps:
@@ -23,8 +25,18 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Get tag from ref
+      - name: Get tag from ref and create if missing
         id: release_tag
         run: |
           TAG="${{ inputs.tag || github.ref_name }}"
+
+          # If the tag doesn't exist, create it on HEAD of main
+          if ! git rev-parse "$TAG" >/dev/null 2>&1; then
+            echo "Tag $TAG does not exist, creating it on main"
+            git config user.name "github-actions[bot]"
+            git config user.email "github-actions[bot]@users.noreply.github.com"
+            git tag "$TAG"
+            git push origin "$TAG"
+          fi
+
           echo "tag=$TAG" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ on:
       tag:
         required: true
         type: string
-        description: "The release tag to publish (must already exist)"
+        description: "The release tag to publish (will be created on main if it doesn't exist)"
 
 jobs:
   setup:


mcowger:
**Claude finished @mcowger's task in 3m 45s** —— [View job](https://github.com/mcowger/plexus/actions/runs/24944449345)

---
### PR Review: Auto-create release tag if it doesn't exist

- [x] Gather context and read changed files
- [x] Analyze for bugs, logic errors, and security issues
- [x] Check test coverage and style concerns
- [x] Post review findings

---

Overall this is a clean, focused change that solves the problem described in #295. The approach is sound. I have a few findings:

### Bug: Tag created on checkout HEAD, not necessarily `main`

In `release-setup.yml:33-38`, when the tag doesn't exist, it's created on whatever `HEAD` the checkout resolved to. Since `actions/checkout@v6` on `workflow_dispatch` defaults to the branch that triggered the workflow (or the default branch), this will usually be `main` — but it's implicit. If someone triggers this workflow from a non-default branch, the tag would point at that branch's HEAD instead. Consider checking out `main` explicitly or at least documenting that the tag is created on the current checkout ref:

```yaml
- uses: actions/checkout@v6
  with:
    fetch-depth: 0
    ref: main  # Ensure tag is created on main, not an arbitrary ref
```

[Fix this →](https://claude.ai/code?q=In%20.github%2Fworkflows%2Frelease-setup.yml%2C%20add%20%60ref%3A%20main%60%20to%20the%20checkout%20step%20so%20that%20auto-created%20tags%20always%20point%20at%20main%20instead%20of%20whatever%20ref%20the%20workflow%20was%20triggered%20on.&repo=mcowger/plexus)

### Potential issue: `git rev-parse "$TAG"` is ambiguous for tag names that look like refs/paths

The `git rev-parse "$TAG"` check on line 34 could match a branch name or remote ref instead of the intended tag. Using `git rev-parse "refs/tags/$TAG"` would be more precise — it only matches actual tags, avoiding a false positive if a branch with the same name exists:

```bash
if ! git rev-parse "refs/tags/$TAG" >/dev/null 2>&1; then
```

[Fix this →](https://claude.ai/code?q=In%20.github%2Fworkflows%2Frelease-setup.yml%20line%2034%2C%20change%20%60git%20rev-parse%20%22%24TAG%22%60%20to%20%60git%20rev-parse%20%22refs%2Ftags%2F%24TAG%22%60%20so%20the%20existence%20check%20only%20matches%20actual%20tags%2C%20not%20branches%20or%20other%20refs%20with%20the%20same%20name.&repo=mcowger/plexus)

### Downstream note: `release-generate-notes.yml` will work correctly

I checked the `Get previous release tag` step in `release-generate-notes.yml:30-52`. It uses `git rev-parse "${CURRENT_TAG}^{}"` which dereferences the tag to a commit — this works fine whether the tag already existed or was just created, since the newly-pushed tag points to a real commit. No issue here.

### Minor: `contents: write` permission scope is broader than needed

The `contents: write` permission on `release-setup.yml:19` is necessary for pushing tags, so this is correct. Just noting it for awareness — it allows writing any repository content, not just tags. There's no narrower GitHub permission for tag-only writes, so this is the best available option.

### Style: No validation on the tag format

When triggered via `workflow_dispatch`, the `tag` input is a freeform string. The push-trigger path constrains tags to the CalVer pattern (`[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9][0-9].[0-9]+`). Consider adding a validation step before creating the tag to prevent typos (e.g., `2026.4.26` instead of `2026.04.26.1`). This is minor since the tag would still work functionally.

### Security: No concerns

The tag push uses `github-actions[bot]` credentials, which is standard practice. The `REPO_SUPER_TOKEN` (seen in other workflow commits) is not used here — the `GITHUB_TOKEN` with `contents: write` is the right scope for tag operations.

---

**Summary:** Two actionable items — pin the checkout ref to `main` to avoid tagging the wrong branch, and use `refs/tags/$TAG` for the existence check. The rest is clean and correct.

PR #298, opened 2026-04-26
feat: add E2E testing infrastructure with mockttp record/replay

### LINKED ISSUE #297: Plexus Lacks E2E Testing
Review the design below and open a PR that implements the code required to make it work. You do not need to record any traces 

# Plexus End-to-End Testing with mockttp Record/Replay

## Overview

This document describes a three-component testing strategy for Plexus that allows deep end-to-end testing without hitting real upstream AI APIs. The approach uses [mockttp](https://github.com/httptoolkit/mockttp) to intercept, record, and replay HTTP traffic — including SSE streaming responses — giving you deterministic, cost-free test runs after an initial recording pass.

The three components are:

1. **Recorder** — runs once against real APIs, captures fixtures to disk
1. **Fixture Server** — runs in tests, serves recorded fixtures as a fake upstream
1. **Request Driver** — the test harness that sends requests into Plexus and asserts on output

-----

## Prerequisites

```bash
bun add --dev mockttp
```

Bun runs TypeScript natively so no separate build step or `ts-node` is needed. No `@types/node` is required either — Bun ships its own type definitions.

Mockttp requires a CA certificate for HTTPS interception. Generate one once and commit it:

```bash
openssl req -x509 -new -nodes \
  -keyout test/fixtures/testCA.key \
  -sha256 -days 3650 \
  -out test/fixtures/testCA.pem \
  -subj '/CN=Plexus Test CA - DO NOT TRUST'
```

**Never trust this certificate system-wide.** It is only for use by the test process via `NODE_EXTRA_CA_CERTS` (Bun respects this environment variable).

-----

## Fixture Format

Each fixture file is NDJSON (one JSON object per line). The filename encodes enough context to be human-readable and version-control friendly.

### Naming Convention

```
test/fixtures/<provider>/<scenario>.ndjson
```

Examples:

```
test/fixtures/anthropic/basic-completion.ndjson
test/fixtures/openai/streaming-chat.ndjson
test/fixtures/openai/tool-call-roundtrip.ndjson
```

### Record Format

Each line in an NDJSON file is one exchange:

```typescript
interface FixtureRecord {
  // The request Plexus made to the upstream
  request: {
    method: string;
    url: string;
    headers: Record<string, string>;
    body: string; // JSON string
  };
  // The response the upstream returned
  response: {
    status: number;
    headers: Record<string, string>;
    // For non-streaming: the full body as a string
    body?: string;
    // For SSE streaming: array of raw SSE event strings, in order
    // Each string is a complete SSE event: "data: {...}\n\n"
    sseEvents?: string[];
  };
}
```

The distinction between `body` and `sseEvents` is detected by the presence of `content-type: text/event-stream` in the response headers.

-----

## Component 1: The Recorder

The recorder runs mockttp in transparent passthrough mode between Plexus and the real upstream APIs. After traffic flows through, it calls `getSeenRequests()` to retrieve all captured exchanges and serializes them to fixture files.

**When to run:** Only when you need to refresh fixtures — when adding new test scenarios, or when upstream API response shapes change.

**Cost:** Hits real APIs and incurs charges. Run deliberately, not in CI.

### recorder.ts

```typescript
import * as mockttp from 'mockttp';
import * as path from 'path';

const FIXTURES_DIR = path.join(import.meta.dir, '../fixtures');

interface FixtureRecord {
  request: {
    method: string;
    url: string;
    headers: Record<string, string>;
    body: string;
  };
  response: {
    status: number;
    headers: Record<string, string>;
    body?: string;
    sseEvents?: string[];
  };
}

export async function runRecorder(
  scenarioName: string,
  provider: string,
  /**
   * A function that drives traffic through Plexus.
   * Plexus should be configured to use `upstreamProxyUrl` as its upstream base URL.
   */
  driveTraffic: (upstreamProxyUrl: string) => Promise<void>
): Promise<void> {
  const server = mockttp.getLocal({
    https: {
      keyPath: path.join(import.meta.dir, '../fixtures/testCA.key'),
      certPath: path.join(import.meta.dir, '../fixtures/testCA.pem'),
    },
  });

  await server.start();

  // Passthrough: forward everything to the real upstream, but record it
  const rule = await server.anyRequest().thenPassThrough({
    // Strip the Authorization header from fixtures to avoid committing API keys.
    // Plexus will inject the real key at runtime; tests use a dummy key.
    beforeRequest: (req) => ({
      headers: {
        ...req.headers,
        authorization: 'Bearer REDACTED',
      },
    }),
  });

  try {
    // Drive real traffic through Plexus, with Plexus pointed at our proxy
    await driveTraffic(server.url);
  } finally {
    const seen = await rule.getSeenRequests();
    const records: FixtureRecord[] = [];

    for (const req of seen) {
      const response = req.response;
      if (!response) continue;

      const contentType = response.headers['content-type'] ?? '';
      const isSSE = contentType.includes('text/event-stream');

      let record: FixtureRecord;

      if (isSSE) {
        // Parse the buffered SSE body into individual events
        const rawBody = await response.body.getText();
        const sseEvents = parseSSEEvents(rawBody);

        record = {
          request: {
            method: req.method,
            url: req.url,
            headers: normalizeHeaders(req.headers),
            body: await req.body.getText(),
          },
          response: {
            status: response.statusCode,
            headers: normalizeHeaders(response.headers),
            sseEvents,
          },
        };
      } else {
        record = {
          request: {
            method: req.method,
            url: req.url,
            headers: normalizeHeaders(req.headers),
            body: await req.body.getText(),
          },
          response: {
            status: response.statusCode,
            headers: normalizeHeaders(response.headers),
            body: await response.body.getText(),
          },
        };
      }

      records.push(record);
    }

    // Write one NDJSON file per scenario using Bun.write
    const dir = path.join(FIXTURES_DIR, provider);
    await Bun.$`mkdir -p ${dir}`;
    const outPath = path.join(dir, `${scenarioName}.ndjson`);
    await Bun.write(
      outPath,
      records.map((r) => JSON.stringify(r)).join('\n') + '\n'
    );
    console.log(`Wrote ${records.length} fixture(s) to ${outPath}`);

    await server.stop();
  }
}

/** Split a raw SSE body into individual event strings */
function parseSSEEvents(raw: string): string[] {
  // SSE events are separated by double newlines
  return raw
    .split(/\n\n/)
    .map((e) => e.trim())
    .filter((e) => e.length > 0)
    .map((e) => e + '\n\n');
}

/** Remove headers that will vary between runs and shouldn't affect matching */
function normalizeHeaders(
  headers: Record<string, string | string[]>
): Record<string, string> {
  const skip = new Set([
    'x-request-id',
    'cf-ray',
    'date',
    'set-cookie',
    // Add any Plexus-injected request tracing headers here
  ]);
  const out: Record<string, string> = {};
  for (const [k, v] of Object.entries(headers)) {
    if (!skip.has(k.toLowerCase())) {
      out[k] = Array.isArray(v) ? v.join(', ') : v;
    }
  }
  return out;
}
```

### Example Recording Script

```typescript
// scripts/record-fixtures.ts
import { runRecorder } from '../test/support/recorder';

// Record: Anthropic basic completion (non-streaming)
await runRecorder(
  'basic-completion',
  'anthropic',
  async (upstreamUrl) => {
    // Start Plexus with upstreamUrl as the Anthropic base URL
    // then send one representative request through it
    const response = await fetch('http://localhost:3000/v1/messages', {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
        'x-api-key': 'test-key',
      },
      body: JSON.stringify({
        model: 'claude-sonnet-4-6',
        max_tokens: 100,
        messages: [{ role: 'user', content: 'Say hello.' }],
      }),
    });
    await response.json();
  }
);

// Record: OpenAI streaming chat
await runRecorder(
  'streaming-chat',
  'openai',
  async (upstreamUrl) => {
    const response = await fetch('http://localhost:3000/v1/chat/completions', {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
        Authorization: 'Bearer test-key',
      },
      body: JSON.stringify({
        model: 'gpt-4o',
        stream: true,
        messages: [{ role: 'user', content: 'Say hello.' }],
      }),
    });
    // Drain the stream
    for await (const _ of response.body!) {}
  }
);
```

-----

## Component 2: The Fixture Server

The fixture server loads NDJSON fixture files and registers mockttp rules that replay them. It runs as Plexus’s upstream during tests.

Key decisions:

- **Matching** is done on method + normalized URL path + request body hash. Headers are excluded (they vary per run due to auth tokens, tracing IDs etc.).
- **SSE replay** uses a `ReadableStream` to drip events back chunk by chunk, exercising Plexus’s streaming pass-through code.
- **Unmatched requests** return 500 immediately and log the unmatched request details, making fixture gaps easy to diagnose.

### fixture-server.ts

```typescript
import * as mockttp from 'mockttp';
import * as path from 'path';
import { readdirSync } from 'fs';

interface FixtureRecord {
  request: {
    method: string;
    url: string;
    headers: Record<string, string>;
    body: string;
  };
  response: {
    status: number;
    headers: Record<string, string>;
    body?: string;
    sseEvents?: string[];
  };
}

type FixtureMap = Map<string, FixtureRecord>;

export class FixtureServer {
  private server: mockttp.Mockttp;
  private fixtures: FixtureMap = new Map();

  constructor() {
    this.server = mockttp.getLocal({
      https: {
        keyPath: path.join(import.meta.dir, '../fixtures/testCA.key'),
        certPath: path.join(import.meta.dir, '../fixtures/testCA.pem'),
      },
    });
  }

  /** Load all fixtures from a directory (recursively) */
  async loadFixtures(dir: string): Promise<void> {
    const files = readdirSync(dir, { recursive: true }) as string[];
    for (const file of files) {
      if (!file.endsWith('.ndjson')) continue;
      const fullPath = path.join(dir, file);
      const text = await Bun.file(fullPath).text();
      const lines = text.split('\n').filter((l) => l.trim());

      for (const line of lines) {
        const record: FixtureRecord = JSON.parse(line);
        const key = this.fixtureKey(
          record.request.method,
          record.request.url,
          record.request.body
        );
        this.fixtures.set(key, record);
      }
    }
    console.log(`Loaded ${this.fixtures.size} fixture(s)`);
  }

  async start(): Promise<string> {
    await this.server.start();

    // Single catch-all rule: match every request against the fixture map
    await this.server.anyRequest().thenCallback(async (req) => {
      const body = await req.body.getText();
      const key = this.fixtureKey(req.method, req.url, body);
      const fixture = this.fixtures.get(key);

      if (!fixture) {
        console.error(
          `[FixtureServer] No fixture for: ${req.method} ${req.url}\n` +
          `  Body hash: ${bodyHash(body)}\n` +
          `  Body: ${body.slice(0, 200)}`
        );
        return {
          status: 500,
          body: JSON.stringify({ error: 'No fixture found for this request' }),
          headers: { 'content-type': 'application/json' },
        };
      }

      const isSSE = fixture.response.sseEvents !== undefined;

      if (isSSE) {
        return this.buildSSEResponse(fixture);
      } else {
        return {
          status: fixture.response.status,
          headers: fixture.response.headers,
          body: fixture.response.body ?? '',
        };
      }
    });

    return this.server.url;
  }

  async stop(): Promise<void> {
    await this.server.stop();
  }

  private buildSSEResponse(fixture: FixtureRecord): mockttp.CallbackResponseResult {
    const events = fixture.response.sseEvents!;
    const encoder = new TextEncoder();

    const stream = new ReadableStream({
      async start(controller) {
        for (const event of events) {
          controller.enqueue(encoder.encode(event));
          // Optional: add a small delay to simulate token drip.
          // Remove or set to 0 for maximum test speed.
          // await Bun.sleep(5);
        }
        controller.close();
      },
    });

    return {
      status: fixture.response.status,
      headers: {
        ...fixture.response.headers,
        'content-type': 'text/event-stream',
        'cache-control': 'no-cache',
        connection: 'keep-alive',
      },
      body: stream,
    };
  }

  /**
   * Compute a stable key for fixture lookup.
   * Uses method + URL path (no host, since the host will differ between
   * recording and replay) + a hash of the normalized request body.
   */
  private fixtureKey(method: string, url: string, body: string): string {
    const parsed = new URL(url);
    const pathAndQuery = parsed.pathname + parsed.search;
    const hash = bodyHash(normalizeRequestBody(body));
    return `${method.toUpperCase()}:${pathAndQuery}:${hash}`;
  }
}

/** Hash a request body for use in fixture lookup keys */
function bodyHash(body: string): string {
  const hasher = new Bun.CryptoHasher('sha256');
  hasher.update(body);
  return hasher.digest('hex').slice(0, 16);
}

/**
 * Normalize a request body before hashing, to make matching stable
 * across runs even if non-semantic fields differ.
 *
 * For AI API requests, fields like `user` tracking IDs or `metadata`
 * should be stripped before hashing if they shouldn't affect which
 * fixture is returned.
 */
function normalizeRequestBody(body: string): string {
  try {
    const parsed = JSON.parse(body);
    // Remove fields that vary per-run but shouldn't affect fixture matching
    delete parsed.user;
    delete parsed.metadata?.request_id;
    // Normalize stream flag -- record both streaming and non-streaming
    // variants separately, or strip it to unify them (your choice)
    // delete parsed.stream;
    return JSON.stringify(parsed, Object.keys(parsed).sort());
  } catch {
    // Not JSON -- use raw body
    return body;
  }
}
```

-----

## Component 3: The Request Driver

The request driver is your test suite. It starts Plexus pointed at the fixture server, sends requests into Plexus’s public API, and asserts on what comes back.

This is where the actual value lies: you’re asserting on Plexus’s *translated, routed, transformed* output — not on the fixture server’s output directly.

### Example Test File (Bun test runner)

```typescript
// test/e2e/anthropic-passthrough.test.ts
import { describe, it, beforeAll, afterAll, expect } from 'bun:test';
import { FixtureServer } from '../support/fixture-server';
import { startPlexus, stopPlexus } from '../support/plexus-harness';
import * as path from 'path';

describe('Anthropic passthrough', () => {
  let fixtureServer: FixtureServer;
  let plexusUrl: string;

  beforeAll(async () => {
    // Start fixture server with all anthropic fixtures
    fixtureServer = new FixtureServer();
    await fixtureServer.loadFixtures(
      path.join(import.meta.dir, '../fixtures/anthropic')
    );
    const upstreamUrl = await fixtureServer.start();

    // Start Plexus with the fixture server as its Anthropic upstream
    plexusUrl = await startPlexus({
      anthropicBaseUrl: upstreamUrl,
      // Use a dummy key -- fixture server ignores auth
      anthropicApiKey: 'test-key-dummy',
    });
  });

  afterAll(async () => {
    await stopPlexus();
    await fixtureServer.stop();
  });

  it('translates an Anthropic messages request end-to-end', async () => {
    const response = await fetch(`${plexusUrl}/v1/messages`, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
        'x-api-key': 'test-key',
      },
      body: JSON.stringify({
        model: 'claude-sonnet-4-6',
        max_tokens: 100,
        messages: [{ role: 'user', content: 'Say hello.' }],
      }),
    });

    expect(response.status).toBe(200);
    const body = await response.json();
    expect(body.type).toBe('message');
    expect(body.content[0].type).toBe('text');
    expect(typeof body.content[0].text).toBe('string');
  });

  it('streams an Anthropic response correctly', async () => {
    const response = await fetch(`${plexusUrl}/v1/messages`, {
      method: 'POST',
      headers: {
        'Content-Type': 'application/json',
        'x-api-key': 'test-key',
      },
      body: JSON.stringify({
        model: 'claude-sonnet-4-6',
        max_tokens: 100,
        stream: true,
        messages: [{ role: 'user', content: 'Say hello.' }],
      }),
    });

    expect(response.status).toBe(200);
    expect(response.headers.get('content-type')).toContain('text/event-stream');

    const events: string[] = [];
    const reader = response.body!.getReader();
    const decoder = new TextDecoder();

    while (true) {
      const { done, value } = await reader.read();
      if (done) break;
      events.push(decoder.decode(value));
    }

    // Assert on the structure of the event stream Plexus produces
    const dataLines = events
      .join('')
      .split('\n')
      .filter((l) => l.startsWith('data: '));

    expect(dataLines.length).toBeGreaterThan(0);
    // Verify Plexus emits message_start before content deltas
    const firstEvent = JSON.parse(dataLines[0].replace('data: ', ''));
    expect(firstEvent.type).toBe('message_start');
  });
});
```

### Plexus Harness (plexus-harness.ts)

This thin wrapper starts and stops a Plexus instance with test-specific config using Bun’s subprocess API:

```typescript
import type { Subprocess } from 'bun';

let plexusProcess: Subprocess | null = null;

export async function startPlexus(config: {
  anthropicBaseUrl?: string;
  openaiBaseUrl?: string;
  anthropicApiKey?: string;
  openaiApiKey?: string;
  port?: number;
}): Promise<string> {
  const port = config.port ?? 0; // 0 = OS assigns a free port
  const env = {
    ...process.env,
    NODE_EXTRA_CA_CERTS: 'test/fixtures/testCA.pem',
    PLEXUS_PORT: String(port),
    ANTHROPIC_BASE_URL: config.anthropicBaseUrl ?? 'https://api.anthropic.com',
    OPENAI_BASE_URL: config.openaiBaseUrl ?? 'https://api.openai.com',
    ANTHROPIC_API_KEY: config.anthropicApiKey ?? 'test-dummy',
    OPENAI_API_KEY: config.openaiApiKey ?? 'test-dummy',
  };

  // Bun runs TypeScript directly -- no build step needed
  plexusProcess = Bun.spawn(['bun', 'run', 'src/server.ts'], {
    env,
    stdout: 'pipe',
    stderr: 'pipe',
  });

  // Wait for Plexus to log its ready message
  return new Promise((resolve, reject) => {
    const timeout = setTimeout(
      () => reject(new Error('Plexus startup timeout')),
      10000
    );

    const reader = plexusProcess!.stdout.getReader();
    const decoder = new TextDecoder();

    async function readLoop() {
      while (true) {
        const { done, value } = await reader.read();
        if (done) break;
        const text = decoder.decode(value);
        // Adjust this to match Plexus's actual startup log line
        const match = text.match(/Listening on port (\d+)/);
        if (match) {
          clearTimeout(timeout);
          resolve(`http://localhost:${match[1]}`);
          return;
        }
      }
    }

    readLoop().catch(reject);
  });
}

export async function stopPlexus(): Promise<void> {
  if (plexusProcess) {
    plexusProcess.kill();
    await plexusProcess.exited;
    plexusProcess = null;
  }
}
```

-----

### First time / refreshing fixtures

```bash
# Set real API keys in environment
export ANTHROPIC_API_KEY=sk-ant-...
export OPENAI_API_KEY=sk-...

# Run the recording script (Bun runs TypeScript directly, no build needed)
NODE_EXTRA_CA_CERTS=test/fixtures/testCA.pem \
  bun run scripts/record-fixtures.ts

# Inspect and commit the fixture files
git add test/fixtures/
git commit -m "chore: refresh API fixtures"
```

### Every test run (CI and local)

```bash
bun test
# No API keys needed
# No network access to upstream APIs
```

-----

### Matching stability

The biggest operational risk is fixture mismatches – a request comes in during tests that doesn’t match any fixture. This is almost always caused by Plexus adding or changing fields in the request body before forwarding upstream. The solution is `normalizeRequestBody()` in the fixture server: strip or canonicalize any fields that Plexus injects dynamically (request tracing IDs, timestamps, etc.) before hashing.

When a mismatch occurs, the fixture server logs the full unmatched request body so you can see exactly what needs to be added to `normalizeRequestBody()` or recorded as a new fixture.

### SSE replay fidelity

The fixture server replays SSE events as a real stream, not as a buffered response. This means Plexus’s streaming pass-through code is genuinely exercised. If you want to test backpressure behavior, add a delay between events in `buildSSEResponse()` using `await Bun.sleep(ms)`. For maximum CI speed, remove the delay entirely.

### Fixture hygiene

- Strip API keys in the recorder’s `beforeRequest` hook (already shown above)
- Strip any PII from message content before committing fixtures – or use synthetic prompts specifically designed for testing (“Say hello.”, “Count to three.”, etc.)
- Fixtures are just JSON – review them in pull requests like any other code

### Multi-provider tests

Load fixture directories selectively per test file:

```typescript
// Only load OpenAI fixtures for OpenAI tests
await fixtureServer.loadFixtures(path.join(import.meta.dir, '../fixtures/openai'));
```

Or load everything and let the key-based lookup handle routing:

```typescript
await fixtureServer.loadFixtures(path.join(import.meta.dir, '../fixtures'));
```

### What this tests vs. what it doesn’t

**Tested by this approach:**

- Plexus’s request translation (e.g. OpenAI format → Anthropic format)
- Plexus’s response translation
- Routing logic (which provider does this model route to?)
- Quota enforcement (does Plexus reject over-quota requests before they hit upstream?)
- Failover logic (configure two fixture servers, make one return 500s)
- Streaming pass-through correctness
- Auth header handling

**Not tested by this approach:**

- Whether upstream APIs have changed their response format (that’s what refreshing fixtures detects)
- Rate limit handling (upstream returns 429) – add a fixture that returns 429 for this
- Network timeout handling – mockttp can simulate delays: `thenCallback(() => Bun.sleep(30000).then(() => ({ status: 200 })))`

mcowger:
## Summary

Implements the E2E testing infrastructure described in #297, using mockttp to intercept, record, and replay HTTP traffic — including SSE streaming responses — for deterministic, cost-free test runs.

### Three-component architecture

1. **Recorder** (`test/support/recorder.ts`) — Runs mockttp as a transparent passthrough proxy to capture real API traffic into NDJSON fixture files. Strips authorization headers automatically.

2. **Fixture Server** (`test/support/fixture-server.ts`) — Loads NDJSON fixture files and replays them as a fake upstream. Matches requests by method + URL path + body hash. SSE responses are replayed as real streams via ReadableStream to exercise Plexus's streaming pass-through code.

3. **Plexus Harness** (`test/support/plexus-harness.ts`) — Starts a Plexus subprocess with test-specific config, waits for startup, then configures providers and model aliases via the management API.

### Files added

- `test/support/fixture-types.ts` — Shared fixture types and matching utilities
- `test/support/recorder.ts` — Component 1: traffic recorder
- `test/support/fixture-server.ts` — Component 2: fixture replay server
- `test/support/plexus-harness.ts` — Component 3: Plexus subprocess manager
- `test/e2e/anthropic-passthrough.test.ts` — Anthropic E2E test (basic + streaming)
- `test/e2e/openai-passthrough.test.ts` — OpenAI E2E test (basic + streaming)
- `test/fixtures/anthropic/*.ndjson` — Sample Anthropic fixtures
- `test/fixtures/openai/*.ndjson` — Sample OpenAI fixtures
- `test/fixtures/testCA.{key,pem}` — CA certificate for mockttp HTTPS interception
- `scripts/record-fixtures.ts` — Script for recording new fixtures against real APIs
- `package.json` — Added `mockttp` dev dependency

Closes #297

## Test plan

- [x] All 1025 existing backend tests pass
- [x] TypeScript type checking passes across all packages
- [x] Biome formatting and linting pass
- [x] Pre-commit hooks (format, lint, typecheck, test) all pass
- [ ] E2E tests require Plexus to be running — verify with `bun test test/e2e/` after starting Plexus
- [ ] Record real fixtures with `NODE_EXTRA_CA_CERTS=test/fixtures/testCA.pem bun run scripts/record-fixtures.ts` and verify E2E tests pass against real Plexus

🤖 Generated with [Claude Code](https://claude.ai/code)


mcowger:
**Claude finished @mcowger's task in 3m 32s** —— [View job](https://github.com/mcowger/plexus/actions/runs/24947232597)

---
### PR Review: E2E Testing Infrastructure with mockttp Record/Replay

- [x] Read project guidelines (AGENTS.md)
- [x] Review fixture types and matching utilities
- [x] Review recorder component
- [x] Review fixture server component
- [x] Review Plexus harness component
- [x] Review E2E test files
- [x] Review record-fixtures script
- [x] Review package.json and fixture files
- [x] Post final review

---

### Architecture & Design

The three-component architecture (Recorder → Fixture Server → Plexus Harness) is clean and well-separated. The record/replay pattern is the right approach for deterministic E2E tests against LLM APIs. Good work overall.

---

### Bugs & Logic Errors

**1. `normalizeHeaders` case mismatch — `SKIP_HEADERS` uses lowercase but comparison is against raw header keys**

`fixture-types.ts:39` — `SKIP_HEADERS` contains lowercase keys (`'x-request-id'`, `'cf-ray'`, etc.), and the comparison does `k.toLowerCase()`. This works correctly. However, the *normalized output* uses the original casing of the input headers. Since mockttp may return headers in various casings, the fixture record could have different header casing than what the fixture server sends back — this isn't a bug per se, but it means response header casing in replay may not match the original recording.

**2. `plexus-harness.ts:60-66` — PORT=0 handling is contradictory**

```ts
if (port !== 0) {
  env.PORT = String(port);
}
// When PORT=0, Plexus defaults to 4000. We'll detect the actual port
// from the startup message. If you need a truly random port, Plexus
// needs to be patched...
if (port === 0) {
  env.PORT = '0';
}
```

The first `if` skips setting PORT when `port === 0`, then the second `if` sets it to `'0'` anyway. The comment says "Plexus defaults to 4000" but the code sets `PORT=0`. If Plexus interprets `PORT=0` literally (it parses `parseInt(process.env.PORT, 10)` → `0` and passes it to `fastify.listen({ port: 0 })`), Fastify will assign a random port. The regex `Server starting on port (\d+)` would then match the **actual** assigned port, not 0. So this likely works, but the comment is misleading and the two `if` blocks are redundant/confusing. The first `if` block is dead code when `port === 0`.

**3. `plexus-harness.ts:68` — Module-level mutable state will break if `startPlexus` is called twice in different test suites**

```ts
let plexusProcess: Subprocess | null = null;
let plexusUrl: string | null = null;
```

These are module-level singletons. The guard at line 37 (`if (plexusProcess) throw new Error(...)`) prevents running two Plexus instances simultaneously, but more importantly, if two test files import this module in the same Bun process, they share the same state. If one test file's `afterAll` calls `stopPlexus()` before the other is done, it will kill the other's Plexus instance. Running `bun test test/e2e/` with both test files would be problematic.

**4. `fixture-server.ts:92-98` — SSE events are enqueued synchronously in `start()`, defeating the purpose of streaming**

```ts
const stream = new ReadableStream({
  async start(controller) {
    for (const event of events) {
      controller.enqueue(encoder.encode(event));
    }
    controller.close();
  },
});
```

All SSE events are enqueued in a single `start()` call, which means the entire response body is buffered before any data is sent. Real SSE servers drip events over time. While the stream still *works* (Plexus will read chunks as they become available from the ReadableStream), this doesn't actually test Plexus's ability to handle incremental/streaming delivery — the entire response body is available immediately. To truly exercise streaming, you'd want to use `pull()` or add small delays between `enqueue()` calls.

**5. `recorder.ts:47` — `rule.getSeenRequests()` may return incomplete data if called immediately after `driveTraffic` resolves**

The `finally` block calls `rule.getSeenRequests()` right after `driveTraffic()` finishes. If the mockttp proxy hasn't fully flushed the response body by that point, `response.body.getText()` could return an incomplete result. This is a race condition — unlikely but possible, especially for large streaming responses.

---

### Security Concerns

**6. Test CA private key committed to the repository**

`test/fixtures/testCA.key` — The private key for the test CA is committed in plaintext. The certificate is explicitly labeled "DO NOT TRUST", so this is clearly a test-only key, but it's still a private key in source control. This is a minor concern since the cert is self-signed and explicitly not for production use, but some security scanners will flag this. Consider adding a `.gitignore` comment or a note in the fixture README.

Also, the `x-request-id` header in the Anthropic fixture (`"x-request-id":"req_test"`) was NOT stripped by `normalizeHeaders` because the `SKIP_HEADERS` check uses `k.toLowerCase()` and the fixture was captured with the key in lowercase — but this header should have been stripped since it's in SKIP_HEADERS. Looking closer: `normalizeHeaders` is called on the *captured* headers, so at recording time, `x-request-id` should have been filtered. The fixture contains it because it was NOT passed through `normalizeHeaders` in the recorder — **wait, actually it IS passed through** (line 67). The issue is that mockttp might return the header key as `X-Request-Id` (title-cased), and `k.toLowerCase()` would convert it to `x-request-id`, which IS in SKIP_HEADERS. Looking at the fixture: the key is `"x-request-id"` in lowercase, so it should be matched. But the value `"req_test"` is present in the fixture, which means `normalizeHeaders` didn't strip it during recording. This means either:
- The recording was hand-crafted (not actually run through the recorder), or
- The SKIP_HEADERS list wasn't applied correctly

This is a minor inconsistency but worth noting.

---

### Missing Test Coverage & Improvements

**7. No test for unmatched requests (fixture miss)**

If Plexus makes a request that doesn't match any fixture, the fixture server returns a 500 with a JSON error. There's no E2E test verifying this behavior or verifying that Plexus handles upstream errors gracefully.

**8. No test for concurrent requests or multiple fixtures in a single scenario**

Each test loads only the fixtures for its provider, and each test makes exactly one request. There's no test exercising multiple sequential requests within the same Plexus instance (e.g., a conversation with follow-up).

**9. No cleanup of the temp database directory**

`plexus-harness.ts:45` — `fs.mkdtempSync` creates a temp directory for the SQLite database, but `stopPlexus()` only kills the process — it never cleans up `tmpDir`. Over many test runs this leaks temp directories.

**10. `bun:test` vs `vitest` inconsistency**

The E2E tests use `bun:test` imports while all other tests in the project use Vitest per `AGENTS.md`. This is intentional (the E2E tests run separately via `bun test test/e2e/`), but there's no script in `package.json` to run the E2E tests. The PR description mentions `bun test test/e2e/` but this should be formalized as an npm script like `"test:e2e"`.

**11. Missing `test:e2e` script in package.json**

No `test:e2e` script was added to `package.json`. Users need a documented, easy way to run the E2E tests.

---

### Minor / Style

**12. `fixture-types.ts:39`** — `SKIP_HEADERS.has(k.toLowerCase())` lowercases the key for lookup, but the normalized output preserves the original key casing. If two headers differ only in casing (e.g., `Content-Type` vs `content-type`), they'd both survive normalization but produce different keys. This is unlikely with real HTTP headers but worth noting.

**13. `recorder.ts:98`** — The final newline after the last JSON record (`+ '\n'`) is good practice for NDJSON, but `parseSSEEvents` in `fixture-types.ts:52` uses `split(/\n\n/)` which splits on double-newline. If the SSE body contains trailing newlines, the final empty event gets filtered by `.filter((e) => e.length > 0)`, so this is handled correctly.

**14. `plexus-harness.ts:53`** — `DEBUG: ''` suppresses all debug logging. This is fine for test output cleanliness but makes debugging test failures harder. Consider using an env var like `PLEXUS_E2E_DEBUG` to optionally re-enable it.

---

### Summary

The architecture is solid and the implementation is mostly clean. The most significant issues are:

| Priority | Issue | Location |
|----------|-------|----------|
| **High** | Module-level singleton state will break when both E2E test files run together | `plexus-harness.ts:6-7` |
| **Medium** | SSE streaming isn't actually tested incrementally — all events are queued at once | `fixture-server.ts:92-98` |
| **Medium** | Redundant/contradictory PORT=0 handling with misleading comment | `plexus-harness.ts:57-66` |
| **Medium** | Temp directory leak (no cleanup in stopPlexus) | `plexus-harness.ts:45` |
| **Low** | Missing `test:e2e` npm script | `package.json` |
| **Low** | Fixture header inconsistency (x-request-id not stripped) | fixture files |

scripts/record-fixtures.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Recording script for E2E test fixtures.
+ *
+ * Run this script with real API keys to capture upstream API responses
+ * into NDJSON fixture files. These fixtures are then replayed by the
+ * FixtureServer during E2E tests, enabling deterministic, cost-free test runs.
+ *
+ * Usage:
+ *   NODE_EXTRA_CA_CERTS=test/fixtures/testCA.pem \
+ *     ANTHROPIC_API_KEY=sk-ant-... \
+ *     bun run scripts/record-fixtures.ts
```
mcowger, line 11: @claude I don't like this pass through the keys as received from plexus, and simply don't record them

mcowger (commented):
Overall this seems to be anthropic specific it should not be.  @claude

scripts/record-fixtures.ts
mcowger: **Claude finished @mcowger's task in 4m 4s** —— [View job](https://github.com/mcowger/plexus/actions/runs/24947384580)

---
### Done: Auth headers now stripped entirely instead of redacted

Made these changes:

1. **`test/support/fixture-types.ts`** — Added `authorization` and `x-api-key` to `SKIP_HEADERS` so they are completely omitted from recorded fixtures (not replaced with a redacted placeholder).

2. **`test/support/recorder.ts`** — Removed the `beforeRequest` hook that was replacing `authorization` with `'Bearer REDACTED'`. This had two problems: it would break real passthrough auth (replacing the real key before forwarding upstream), and it didn't strip `x-api-key` at all. Now the proxy is a clean passthrough — real keys reach upstream APIs, but `normalizeHeaders` strips them from the fixture output.

3. **Fixture files** — Removed `"x-api-key":"REDACTED"` and `"authorization":"Bearer REDACTED"` from all 4 fixture NDJSON files.

All validation passes (typecheck, format, 1025 backend tests).

mcowger:
**Claude finished @mcowger's task in 1m 48s** —— [View job](https://github.com/mcowger/plexus/actions/runs/24947452789)

---
### PR Review: E2E Testing Infrastructure with mockttp Record/Replay

- [x] Read project guidelines (AGENTS.md)
- [x] Review fixture types and matching utilities
- [x] Review recorder component
- [x] Review fixture server component
- [x] Review Plexus harness component
- [x] Review E2E test files
- [x] Review record-fixtures script
- [x] Review fixture files and CA cert
- [x] Post final review

---

The three-component architecture (Recorder → Fixture Server → Plexus Harness) is clean and well-separated. The record/replay pattern is the right approach for deterministic E2E tests. Below are the substantive issues found.

---

### Bugs & Logic Errors

**1. Module-level singleton state breaks when both E2E test files run together — HIGH**

`plexus-harness.ts:6-7` — `plexusProcess` and `plexusUrl` are module-level singletons. If `bun test test/e2e/` runs both test files in the same process, the second file's `beforeAll` will hit the guard at line 37 and throw. More critically, if the first file's `afterAll` calls `stopPlexus()` while the second is still running, it kills the other's instance.

[Fix this →](https://claude.ai/code?q=Refactor%20plexus-harness.ts%20to%20use%20a%20class%20instead%20of%20module-level%20singletons%20so%20each%20test%20suite%20gets%20its%20own%20Plexus%20instance.%20The%20class%20should%20hold%20plexusProcess%20and%20plexusUrl%20as%20instance%20properties.&repo=mcowger/plexus)

**2. SSE streaming is not actually tested incrementally — all events enqueue at once — MEDIUM**

`fixture-server.ts:92-98` — All SSE events are enqueued in a single `start()` call, meaning the entire response body is buffered before any data flows. This doesn't test Plexus's ability to handle incremental streaming delivery. Consider using `pull()` or adding small delays between `enqueue()` calls.

[Fix this →](https://claude.ai/code?q=In%20fixture-server.ts%20buildSSEResponse%2C%20change%20the%20ReadableStream%20from%20using%20start()%20to%20using%20pull()%20with%20small%20delays%20between%20SSE%20events%20to%20exercise%20incremental%20streaming.%20Add%20a%20configurable%20delay%20(default%2010ms)%20between%20enqueue%20calls.&repo=mcowger/plexus)

**3. Redundant PORT=0 handling with misleading comment — MEDIUM**

`plexus-harness.ts:57-66` — The first `if (port !== 0)` skips setting PORT when `port === 0`, then the second `if (port === 0)` immediately sets `env.PORT = '0'`. The two blocks are contradictory: the first implies PORT=0 should be left unset, the comment says Plexus defaults to 4000, but the second block sets it to `'0'` anyway. If Plexus's `parseInt(process.env.PORT, 10)` returns 0 and passes it to Fastify, it gets a random port — in which case the startup regex will catch the real port. Either way, the logic should be simplified to just `env.PORT = String(port)` unconditionally and the comment corrected.

[Fix this →](https://claude.ai/code?q=In%20plexus-harness.ts%2C%20replace%20the%20two%20PORT%20if-blocks%20(lines%2057-66)%20with%20a%20single%20line%20%60env.PORT%20%3D%20String(port)%3B%60%20and%20update%20the%20comment%20to%20accurately%20describe%20the%20PORT%3D0%20behavior.&repo=mcowger/plexus)

**4. Temp directory leak — no cleanup in stopPlexus — MEDIUM**

`plexus-harness.ts:45` — `fs.mkdtempSync` creates a temp directory for the SQLite database, but `stopPlexus()` only kills the process and never cleans up `tmpDir`. Over many test runs this leaks temp directories on disk.

[Fix this →](https://claude.ai/code?q=In%20plexus-harness.ts%2C%20store%20tmpDir%20alongside%20plexusProcess%20and%20in%20stopPlexus()%20call%20fs.rmSync(tmpDir%2C%20%7B%20recursive%3A%20true%20%7D)%20after%20the%20process%20exits.&repo=mcowger/plexus)

**5. `x-request-id` not stripped from fixture files — LOW**

The Anthropic fixtures (`basic-completion.ndjson`, `streaming-completion.ndjson`) still contain `"x-request-id":"req_test"` in the response headers, despite `x-request-id` being in `SKIP_HEADERS`. This happens because `normalizeHeaders` is only called on the *request* headers in the recorder, not the response headers — the response headers are passed through raw from mockttp. The fixture server then replays those headers including `x-request-id`.

[Fix this →](https://claude.ai/code?q=In%20recorder.ts%2C%20apply%20normalizeHeaders%20to%20both%20req.headers%20and%20response.headers%20when%20building%20the%20FixtureRecord.%20Also%20remove%20x-request-id%20from%20the%20existing%20Anthropic%20fixture%20NDJSON%20files.&repo=mcowger/plexus)

---

### Security

**6. Test CA private key committed to the repository — LOW**

`test/fixtures/testCA.key` — The private key for the test CA is in plaintext in source control. The cert is self-signed and explicitly not for production use, so this is minor, but security scanners will flag it. Consider adding a note in the fixture directory or a `.gitignore` exception with explanation.

---

### Missing Coverage & Improvements

**7. No test for unmatched requests (fixture miss) — MEDIUM**

If Plexus makes a request that doesn't match any fixture, the fixture server returns a 500. There's no test verifying that Plexus handles upstream errors gracefully.

**8. No `test:e2e` npm script — LOW**

`package.json` — No script was added for running E2E tests. The PR description mentions `bun test test/e2e/` but this should be formalized as `"test:e2e"` in `package.json`.

**9. `bun:test` vs `vitest` inconsistency — LOW**

The E2E tests import from `bun:test` while all other tests use Vitest per AGENTS.md. This is intentional (the tests run separately), but it means the E2E tests bypass the project's existing test infrastructure (mocking rules, global setup, etc.). Fine for now, but worth noting as the E2E suite grows.

**10. No test for concurrent or sequential requests — LOW**

Each test makes exactly one request. There's no test exercising multiple sequential requests within a Plexus instance (e.g., a conversation with follow-up).

---

### Summary

| Priority | Issue | Location |
|----------|-------|----------|
| **High** | Module-level singleton state breaks when both E2E test files run together | `plexus-harness.ts:6-7` |
| **Medium** | SSE streaming isn't actually tested incrementally — all events queued at once | `fixture-server.ts:92-98` |
| **Medium** | Redundant/contradictory PORT=0 handling with misleading comment | `plexus-harness.ts:57-66` |
| **Medium** | Temp directory leak (no cleanup in stopPlexus) | `plexus-harness.ts:45` |
| **Medium** | No test for unmatched requests (fixture miss) | test files |
| **Low** | `x-request-id` not stripped from fixture response headers | `recorder.ts`, fixtures |
| **Low** | Missing `test:e2e` npm script | `package.json` |
| **Low** | Test CA private key committed to repository | `test/fixtures/testCA.key` |

[View job](https://github.com/mcowger/plexus/actions/runs/24947452789)

mcowger:
@claude a few more items to fix


 Items to fix 

1. The plexus database should be a SQLIte :memory: database so that it doesn’t store data on disk
2. Module-level singleton state breaks when both E2E test files run together should be protected against by forcing them them run sequentially 
3. Redundant/contradictory PORT=0 handling with misleading comment < port 0 is not respected.  Force plexus to a known unused port like 4444
4. x-request-id not stripped from fixture response headers < fix this
5. Missing test:e2e npm script
6. E2E tests should be vitest not bun 
7. Remove the sample ndjson test files.

mcowger:
**Claude finished @mcowger's task in 4m 7s** —— [View job](https://github.com/mcowger/plexus/actions/runs/24948017921)

---
### Fixing requested items 

- [ ] 1. Use SQLite `:memory:` database instead of on-disk
- [ ] 2. Force E2E tests to run sequentially (protect against singleton state)
- [ ] 3. Force Plexus to port 4444 instead of PORT=0 handling
- [ ] 4. Strip `x-request-id` from fixture response headers
- [ ] 5. Add `test:e2e` npm script
- [ ] 6. Convert E2E tests from `bun:test` to `vitest`
- [ ] 7. Remove sample NDJSON fixture files

mcowger (commented):
@claude 

Items to fix

The plexus database should be a SQLIte :memory: database so that it doesn’t store data on disk
Module-level singleton state breaks when both E2E test files run together should be protected against by forcing them them run sequentially
Redundant/contradictory PORT=0 handling with misleading comment < port 0 is not respected. Force plexus to a known unused port like 4444
x-request-id not stripped from fixture response headers < fix this
Missing test:e2e npm script
E2E tests should be vitest not bun
Remove the sample ndjson test files.

mcowger:
Dont like this design.

PR #301, opened 2026-04-26
Update Quota Tracking

mcowger:
Replaces the old QuotaWindow[]-based system with a flat Meter[] model
where each meter is self-describing with kind, unit, period fields, and
status — eliminating per-checker UI switches, checker-level categories,
compound enum window types, and 21 near-duplicate display components.

Backend changes:
- New Meter/MeterCheckResult types in src/types/meter.ts
- Self-registering checker-registry.ts (defineChecker + loadAllCheckers)
- New meter_snapshots table schema (SQLite + PostgreSQL)
- All 22 checkers rewritten to export default defineChecker({...})
- apertis and apertis-coding-plan merged into single apertis checker
- quota-scheduler.ts rewired to use registry + persist to meter_snapshots
- /v0/management/quotas routes return MeterCheckResult shape
- apertis-coding-plan removed from config schema

Frontend changes:
- New QuotaCheckerInfo type with meters: Meter[] replacing latest: snapshot[]
- New generic components: BalanceMeterRow, AllowanceMeterRow, MeterValue
- checker-presentation.ts for display names
- Quotas.tsx, CombinedBalancesCard, CompactQuotasCard, CompactBalancesCard
  rewritten to use meters array directly
- Sidebar.tsx updated to derive balance/allowance from meter kinds
- All 21 XxxQuotaDisplay.tsx components deleted
- QuotaHistoryModal, BalanceHistoryModal deleted
- api.ts updated to use new response shape


mcowger:
**Claude finished @mcowger's task in 8m 51s** —— [View job](https://github.com/mcowger/plexus/actions/runs/24971247892)

---
### PR Review: Update Quota Tracking

- [x] Gather context and read key files
- [x] Review core types and checker registry
- [x] Review backend quota scheduler and routes
- [x] Review frontend components and API changes
- [x] Post final review findings

---

### Overall Assessment

This is a solid architectural refactor. The Meter-based model eliminates 21 near-duplicate display components and a complex per-checker switching system, replacing it with a uniform, self-describing data model. The code is well-structured and the migration path (legacy migrator, CLI subcommand, backup/restore UI) is thorough.

That said, I found several issues ranging from a SQL injection vulnerability to logic bugs and type drift. Details below.

---

### Critical Issues

**1. SQL Injection in `legacy-snapshot-migrator.ts:92`**

`tableExists()` uses `sql.raw()` to interpolate a table name directly into a query string:

```ts
await selectAll(db, sql.raw(`SELECT 1 FROM ${tableName} LIMIT 0`));
```

While `tableName` is currently only called with the hardcoded string `'quota_snapshots'`, `sql.raw()` does **no parameterization** — if this function is ever called with user input, it's a direct SQL injection vector. This should use drizzle's identifier escaping or at minimum validate `tableName` against an allowlist.

[Fix this →](https://claude.ai/code?q=In%20packages%2Fbackend%2Fsrc%2Fservices%2Fquota%2Flegacy-snapshot-migrator.ts%20line%2092%2C%20the%20sql.raw()%20call%20with%20string%20interpolation%20of%20tableName%20is%20a%20SQL%20injection%20risk.%20Either%20add%20an%20allowlist%20check%20for%20known%20table%20names%20or%20use%20drizzle%20identifier%20escaping.%20Branch%3A%20rebase-pr-276)

**2. SQL Injection in `exportLegacySnapshots` — `sqlEscape()` is insufficient**

`legacy-snapshot-migrator.ts:294-300`: The `sqlEscape()` function only escapes single quotes via `replaceAll("'", "''")`. This is used to generate SQL `INSERT` statements. While not directly exploitable from a web endpoint (the output is a downloaded file), if any `quota_snapshots` row ever contains a value with `; DROP TABLE`-style payloads in its string columns, the exported `.sql` file would contain them verbatim. This is a defense-in-depth concern more than a live exploit, but worth noting.

---

### Bugs / Logic Errors

**3. `balance()` in `checker-registry.ts:122-126` — Incorrect utilization derivation**

The `balance()` factory method in `MeterContext` overrides the general `deriveUtilization()` logic:

```ts
balance(params: BalanceParams): Meter {
  const utilization: Utilization =
    params.limit !== undefined && params.used !== undefined
      ? deriveUtilization(params.used, params.limit, params.remaining)
      : 'not_applicable';
```

This means **balance meters with no `limit` and no `used` but with `remaining`** (e.g., a simple PAYG wallet) will get `utilizationPercent: 'not_applicable'` instead of `'unknown'`. This isn't exactly wrong semantically, but it differs from the `allowance()` path which uses `deriveUtilization()` unconditionally. The inconsistency may cause display issues — the `AllowanceMeterRow` renders a progress bar for numbers but shows nothing for `'not_applicable'`, and `BalanceMeterRow` shows "—" for undefined display values derived from `'not_applicable'` utilization.

[Fix this →](https://claude.ai/code?q=In%20packages%2Fbackend%2Fsrc%2Fservices%2Fquota%2Fchecker-registry.ts%20lines%20122-126%2C%20the%20balance()%20method%20returns%20'not_applicable'%20when%20both%20limit%20and%20used%20are%20undefined%2C%20but%20this%20causes%20BalanceMeterRow%20to%20show%20%E2%80%9C%E2%80%94%E2%80%9D%20for%20meters%20that%20have%20remaining%20but%20no%20limit.%20Consider%20using%20'unknown'%20instead%20so%20the%20remaining%20value%20is%20still%20displayed.%20Branch%3A%20rebase-pr-276)

**4. `getLatestQuota()` grouping by exact `checkedAt` ms is fragile**

`quota-scheduler.ts:295`: Grouping the "latest" snapshot rows by exact `checkedAt` millisecond:

```ts
const latestRows = rows.filter((r: any) => toMs(r.checkedAt) === latestMs);
```

If two check cycles for the same checker produce slightly different `checkedAt` values within the same batch (e.g., a multi-meter checker where each meter's persist completes a few ms apart), some meters from the latest check could be excluded. This is unlikely in practice since `checkedAt` is set once before the loop at line 95, but if the persist is retried or overwritten, it could cause missing meters in the response.

**5. Frontend `QuotaConfig` type in `api.ts:928-941` is stale**

The `QuotaConfig` interface in `api.ts` still lists `type` as a union of only ~11 checker types:

```ts
type:
  | 'synthetic' | 'naga' | 'nanogpt' | 'codex' | 'claude-code'
  | 'zai' | 'moonshot' | 'minimax' | 'minimax-coding'
  | 'kimi-code' | 'openrouter' | 'kilo';
```

This is missing at least `copilot`, `wisdomgate`, `apertis`, `ollama`, `gemini-cli`, `antigravity`, `poe`, `neuralwatt`, `zenmux`, and `openai-codex`. It also includes `'codex'` which doesn't appear in the registry (the checker is `'openai-codex'`). While `QuotaConfig` appears unused in the actual API call path (the API returns `QuotaCheckerInfo`), this stale type could confuse future developers or cause TS errors if someone does try to use it.

[Fix this →](https://claude.ai/code?q=In%20packages%2Ffrontend%2Fsrc%2Flib%2Fapi.ts%20lines%20928-941%2C%20the%20QuotaConfig%20type%20union%20for%20%60type%60%20is%20missing%20many%20checker%20types%20(copilot%2C%20wisdomgate%2C%20apertis%2C%20ollama%2C%20gemini-cli%2C%20antigravity%2C%20poe%2C%20neuralwatt%2C%20zenmux%2C%20openai-codex)%20and%20includes%20a%20stale%20%27codex%27%20entry.%20Either%20update%20the%20union%20or%20replace%20it%20with%20%60string%60%20to%20match%20the%20registry-based%20approach.%20Branch%3A%20rebase-pr-276)

---

### Design Concerns

**6. Self-registering checkers via module-level side effects**

`checker-registry.ts:176-199`: The `loadAllCheckers()` function works by importing each checker module, which calls `defineChecker()` at module top-level as a side effect. This is fragile:

- If a checker module is imported anywhere else in the codebase (e.g., for its types), it silently double-registers or overwrites the registry entry.
- Adding a new checker requires editing `loadAllCheckers()` — the registry is not truly automatic.
- The `defineChecker()` call happens at import time, meaning test files that import checker modules directly will pollute the registry.

Consider making `loadAllCheckers()` the single entry point and removing the side-effect registration, or use a dynamic import glob pattern if Bun supports it.

**7. `Meter` type duplicated between backend and frontend**

`packages/backend/src/types/meter.ts` and `packages/frontend/src/types/quota.ts` both define the `Meter` interface independently. The frontend version omits `exhaustionThreshold`. This is a maintenance hazard — any field added to one must be manually mirrored. Consider sharing the type from a common package (e.g., `@plexus/shared`).

**8. `any` casts throughout `quota-scheduler.ts`**

The `getLatestQuota()` and `getQuotaHistory()` methods heavily use `as any[]` and `as any` casts (lines 291, 310, etc.). While this works, it defeats the type safety that drizzle provides. The row type from `schema.meterSnapshots` is fully known at compile time.

---

### Minor Issues

**9. `BalanceMeterRow` doesn't show limit/used information**

`BalanceMeterRow.tsx:12-16`: When `remaining` is undefined but both `used` and `limit` exist, it computes `displayValue = limit - used`. But if `remaining` is `0`, it shows `0` which is indistinguishable from "no data" in some contexts. Consider showing the original `remaining` directly and only computing `limit - used` as a fallback.

**10. Unused `rawCreds` variable in `openai-codex-checker.ts:124-128`**

```ts
const rawCreds = (
  oauthAccountId
    ? authManager.getCredentials(provider as OAuthProvider, oauthAccountId)
    : authManager.getCredentials(provider as OAuthProvider)
) as Record<string, unknown> | null;
```

`rawCreds` is assigned but never read. The credentials are re-fetched on lines 144-148.

**11. Missing `exhaustionThreshold` in frontend `Meter` type**

`packages/frontend/src/types/quota.ts:6-21`: The frontend `Meter` type doesn't include `exhaustionThreshold`, which the backend type has. If the frontend ever needs to render per-meter exhaustion thresholds (e.g., different colored bars per meter), this field will be silently dropped from the API response.

**12. `MeterHistoryModal` redeclares `HistoryRow` interface locally**

`MeterHistoryModal.tsx:38-47`: The `HistoryRow` type is declared locally rather than derived from the backend's meter_snapshots schema. Any schema changes (added columns, renamed fields) won't surface as type errors here.

**13. No database index on `meter_snapshots.success`**

The schema has indexes on `(checker_id, meter_key, checked_at)`, `(provider, checked_at)`, and `(checked_at)`, but the `getLatestQuota()` method filters by `success` as part of its error-path logic. If the table grows large, filtering failed checks might benefit from a partial or composite index.

---

### Test Coverage

- The `quota-scheduler.test.ts` covers persistence, threshold-based cooldowns, and the strict-vs-lenient checker interaction. Good coverage of the cooldown logic.
- The checker-level tests were updated to use the new `Meter` shape and `defineChecker()` API. The deleted `apertis-coding-plan-checker.test.ts` is correct since that checker was merged into `apertis-checker`.
- **Gap**: No tests for `legacy-snapshot-migrator.ts` — this is substantial ETL logic with raw SQL, batch fallback, and dialect-specific paths. A test covering the `mapWindowType()` mapping and a round-trip migration would catch regressions.
- **Gap**: No frontend tests for the new components (`AllowanceMeterRow`, `BalanceMeterRow`, `MeterHistoryModal`, `CompactQuotasCard`). These are pure-presentational so the risk is lower, but `MeterHistoryModal` has non-trivial data-fetching logic.

---

### Summary

| Severity | Count | Key Items |
|----------|-------|-----------|
| Critical | 1 | SQL injection via `sql.raw()` |
| Bug | 4 | Balance utilization inconsistency, stale frontend QuotaConfig type, fragile checkedAt grouping, unused variable |
| Design | 3 | Side-effect registration, duplicated Meter type, excessive `any` casts |
| Minor | 5 | Missing exhaustionThreshold in frontend, no migrator tests, etc. |

The refactor achieves its stated goal of simplifying the quota system. The main items to address before merge are the SQL injection risk (#1) and the stale frontend `QuotaConfig` type (#5). The balance/allowance utilization inconsistency (#3) should also be resolved to avoid display edge cases.

PR #302, opened 2026-04-26
feat(backend): real tokenizer + image-aware token counting

darkspadez:
Replace the JSON.stringify-everything heuristic in estimate-tokens with a
structural walker that routes each text part through tiktoken's o200k_base
encoder and each image through a per-provider formula (Claude (w*h)/750 cap
1600, OpenAI tile-based with low-detail = 85, Gemini fixed 258). Image
dimensions are sniffed from base64 PNG/JPEG/GIF/WebP headers without a full
decode; external URLs and unknown formats fall back to a conservative default.

The old path treated a 1MB base64 image as ~330k heuristic tokens; the new
walker counts the same image at ~1.5k for Claude/OpenAI and 258 for Gemini —
roughly a 200x correction on the dominant source of estimator error.

Heuristic is retained as a labeled fallback for: encoder load failures, single
strings longer than 32k chars (where BPE is O(n^2) on adversarial repetitive
input and the heuristic is accurate enough at scale), and unknown apiType.

The 10% safety multiplier in enforce-limits stays — o200k_base still has
~5-15% drift vs. Claude/Gemini tokenizers, so the existing margin is sized
appropriately for the new tokenizer.


mcowger:
## ⚠️ This PR was reverted — tiktoken WASM is not viable in compiled binaries

This PR was reverted in d97541a7 because **tiktoken crashes the Docker container at startup**.

### Root cause

`tiktoken`'s CJS entrypoint (`tiktoken.cjs`) loads `tiktoken_bg.wasm` via `fs.readFileSync` **at require-time**, walking up the filesystem from `__dirname` to find `node_modules/tiktoken/tiktoken_bg.wasm`. In a compiled Bun binary (`bun build --compile`), `node_modules` does not exist on the real filesystem — all modules are bundled inside the binary — so every candidate path fails, `bytes` stays `null`, and it throws:

```
Error: Missing tiktoken_bg.wasm
```

This kills the process before the HTTP server ever starts, causing the Docker smoke test (and any real deployment) to fail immediately.

### Why this is hard to work around

The throw happens at **module load time**, before any try/catch in application code can intercept it. The existing `getEncoder()` fallback in `estimate-tokens.ts` never gets a chance to run.

Adding the `.wasm` to the `bun build --compile` asset globs does not help either — `fs.readFileSync` on a runtime filesystem path does not consult Bun's embedded asset store.

### What the contributor should do

Please resubmit the token-counting improvement using a **pure-JS tokenizer** with no native/WASM dependency, so it works correctly in both `bun run` (dev) and `bun build --compile` (Docker/production) modes. A few options:

- [`gpt-tokenizer`](https://www.npmjs.com/package/gpt-tokenizer) — pure TypeScript, no WASM, accurate for cl100k / o200k
- Keep the existing heuristic and only land the image-dimension sniffing / per-provider formulas (the non-tiktoken parts of this PR), which were valuable and had no WASM dependency

The image-aware token counting logic itself (provider-specific formulas, base64 dimension sniffing) is good work worth preserving in a follow-up PR.

Issue #304, opened 2026-04-27
meter_snapshots INSERT fails on PostgreSQL: boolean sent for integer "success" column


thangnguyen-69:
**Describe the bug**
On PostgreSQL, every meter snapshot INSERT fails because the `success` column is typed as `integer` in the schema but the quota-scheduler sends a JavaScript boolean (`true`/`false`). PostgreSQL strict type checking rejects the query with `column "success" is of type integer but expression is of type boolean`.

**To Reproduce**
1. Run Plexus with `DATABASE_URL=postgres://...`
2. Wait for any quota check interval, or trigger one via API.
3. Logs show repeated errors like:
```
Failed to persist meter 'energy_quota' for 'neuralwatt': Error: Failed query:
insert into "meter_snapshots" (..., "success", ...) values (...)
params: ...,true,...
```

**Environment**
- Plexus version: `ghcr.io/mcowger/plexus:2026.04.26.3` (latest at time of report)
- Database: PostgreSQL 17 (CloudNativePG)
- Driver: `postgres-js` (default)

**Diagnosis**
Drizzle schema defines `success` as:
```ts
success: integer('success').notNull().default(1),
```
But `quota-scheduler.ts` passes:
```ts
success: true   // or success: false
```
Drizzle for SQLite handles this silently (sqlite auto-casts), but PostgreSQL does not.

**Suggested fixes (pick one)**
1. **Change code** (`quota-scheduler.ts`): Cast boolean to integer before insert:
   ```ts
   success: result.success ? 1 : 0,
   ```
2. **Change schema** (`meter-snapshots.ts`): Use `boolean` type instead:
   ```ts
   success: boolean('success').notNull().default(true),
   ```
   This would also need a follow-up migration. SQLite uses `INTEGER` under the hood for booleans so this is actually the more natural Drizzle idiom.

**Additional context**
- Other columns (`group`, `limit`) are reserved SQL keywords, but Drizzle properly double-quotes them, so those work fine. The issue is exclusively the `integer` vs `boolean` type mismatch on PostgreSQL.
- The pod starts fine and serves traffic — only meter snapshot persistence is affected.

mcowger:
Resolved in #309

PR #309, opened 2026-04-27
fix(quota): use boolean type for meter_snapshots.success on Postgres

darkspadez:
The Postgres schema typed `success` as integer while the quota scheduler inserts a JS boolean, so every meter snapshot INSERT failed on Postgres with `column "success" is of type integer but expression is of type boolean`. SQLite hid this because its column is declared with `integer({ mode: 'boolean' })`, which auto-casts.

Switch the Postgres column to `boolean` (default `true`) so both dialects present a boolean to the application — no callsite changes needed and the existing `!r.success` read paths work for both.


darkspadez:
This should address #304 .

PR #355, opened 2026-05-06
ci: add multi-arch Docker builds (amd64 + arm64)

harrisony:
## Summary

- Add `compile:linux-arm64` script and use `TARGETARCH` in the Dockerfile to select the correct compile target at build time
- Add QEMU and Docker Buildx steps to the release workflow, enabling `platforms: linux/amd64,linux/arm64` multi-arch image builds
- Remove hardcoded `platform: linux/amd64` from `docker-compose.yml` so the platform is auto-detected

## Verification

Multi-arch build tested successfully via `workflow_dispatch`:  
https://github.com/harrisony/plexus/actions/runs/25427061173


mcowger:
How much does this increase the image size?

PR #383, opened 2026-05-13
A small opencode plugin to add plexus as a provider

IndianBoy42:
1. adds plexus to the auth hook so you can do `/connect` or `opencode auth login`
you have to add the provider manually in opencode.jsonc, doesn't seem to be a way around that, but also the api base url has to be specified anyway
```
    "plexus": {
      "api": "http://localhost:4000/v1",
      "name": "Plexus",
    },
```
2. fetches and adds all the models from plexus the selector
3. a second commit, which is optional, does a few tweaks to the `v1/models/` endpoint to add image modalities for alias that has use_image_fallthrough
4. another optional commit, I wonder if it is an okay idea to expose the management/alias endpoint without the admin_key (just normal auth), this would allow getting richer information for things like this plugin, and in case (2) is undesired.

### DIFF
--- a/packages/backend/src/routes/inference/models.ts
+++ b/packages/backend/src/routes/inference/models.ts
@@ -21,6 +21,7 @@ export async function registerModelsRoute(fastify: FastifyInstance) {
     const metadataManager = ModelMetadataManager.getInstance();
 
     const created = Math.floor(Date.now() / 1000);
+    const hasVisionFallthrough = !!config.vision_fallthrough;
 
     const models = Object.entries(config.models).map(([aliasId, modelConfig]) => {
       const metaConfig = modelConfig?.metadata;
@@ -33,6 +34,17 @@ export async function registerModelsRoute(fastify: FastifyInstance) {
       };
 
       if (!metaConfig) {
+        // When vision fallthrough is enabled for this alias, advertise image
+        // input capability even without explicit metadata.
+        if (hasVisionFallthrough && modelConfig.use_image_fallthrough) {
+          return {
+            ...base,
+            architecture: {
+              input_modalities: ['text', 'image'],
+              output_modalities: ['text'],
+            },
+          };
+        }
         return base;
       }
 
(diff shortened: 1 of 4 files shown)


mcowger:
Thanks for putting this together @IndianBoy42! There's some good stuff here but I think this PR is doing too much at once, and one of the changes
 concerns me.

The vision fallthrough modality fix on /v1/models is great. That's exactly the right place to surface that information so downstream consumers (like your plugin) can get what they need without hitting management endpoints. I'd love to see that as its own PR so we can merge it quickly.

The management aliases change is where I'm stuck. Making /v0/management/aliases available to any authenticated user feels like the wrong approach. Those endpoints return internal routing config (providers, selectors, targets) which is admin-level info. If the plugin needs model capabilities, context lengths, modalities, pricing, that's what /v1/models is for. The right fix is to make /v1/models more complete (which your fallthrough commit already does!), not to open up the management API. I don't think we should weaken auth on management routes to work around a gap in the public endpoint.

As for the opencode plugin itself, a 345-line plugin with its own release cadence and maintenance doesn't really belong in scripts/ alongside operational tooling. It should live in its own repo. You could publish it as a standalone project and other opencode users would have an easier time finding it too - I'd be happy to link it from the docs.

 So my ask: could you split this up? The fallthrough modality fix as one PR (that one I'm happy to merge), and we can discuss the other two separately. The plugin definitely deserves its own repo, and I'd want to talk more about the auth question before going that direction.

IndianBoy42:
No problem, I tried my best to split it as commits and pushed as a draft so they could be cherry picked as you want. I wasn't sure if there's a reason the vision fall through information wasn't part of v1/models already. Do you want to cherrypick the commit or I create a pr? 

I'll think about the plugin as its own repo. Feels overkill for a single file (only 205 lines without the third commit) and then I'd have to learn how npm works to actually do it. It doesn't really even need a release cadence because I think it's just done? Only possible improvements I can think would involve expanding the scope unnecessarily

mcowger:
I'll cherry pick the commit.  

For plugin - i get it.  If you don't want to do npm, I think it's probably fine to just host it has a gist.

PR #404, opened 2026-05-15
feat(dashboard): add pie/bar chart toggle to Usage Analytics

micookie2:
## Summary
- Add a per-card toggle button (pie/bar icons) in the card header of each distribution chart, allowing users to switch between pie chart and vertical bar chart
- Data is sorted by value (descending) for both chart types, so the largest category appears first
- Bar chart has no hover shadow effect (`activeBar={false}`)
- Default view is pie chart (preserves current UX), each card toggles independently

## Test plan
- [ ] Open Dashboard → Usage Analytics tab
- [ ] Verify each distribution card (Model/Provider/API Key x Requests/Tokens) shows a toggle button pair in the top-right corner
- [ ] Click the bar chart icon — card switches to vertical bar chart, data sorted descending left-to-right
- [ ] Click the pie chart icon — card switches back to pie chart, data sorted descending
- [ ] Hover over bar chart bars — no shadow/active effect, only tooltip appears
- [ ] Verify toggle state is per-card (switching one card does not affect others)
- [ ] Verify other charts (Requests over Time, Token Usage, Concurrency) are unaffected

### DIFF
--- a/packages/frontend/src/components/dashboard/tabs/UsageTab.tsx
+++ b/packages/frontend/src/components/dashboard/tabs/UsageTab.tsx
@@ -28,6 +28,7 @@ import { useEffect, useMemo, useState } from 'react';
  * Each record represents a single (provider, model, timestamp) data point returned
  * from the `GET /v0/management/concurrency?timeRange=...` endpoint.
  */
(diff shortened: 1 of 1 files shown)


micookie2:
image

PR #413, opened 2026-05-17
feat(frontend): add zh/en i18n frontend

micookie2:
Introduce react-i18next with browser language detection, language switcher, and locale bundles for pages, dashboard, provider/model modals, and quota checker configs. Quota editors use a single useT('quotas') with checkerConfigs.* and checkerCommon.* keys.


mcowger:
Hey, thank you so much for the incredible effort here. I can see the amount of work that went into extracting and translating all ~1,400 keys across the entire frontend. The i18n infrastructure, the locale files, the component refactors. It's all very thorough and well-done.

After sitting with it though, I'm going to decline this one, and I want to explain why honestly.

This is primarily a personal project, and the ongoing cost of this approach gives me real pause. Every new string, label, error message, or component I add in the future requires updating translation files in two (or more) languages. That's ~1,700 keys of domain-specific content (model architectures, quota configs, provider settings) that need to stay in sync, and that tax doesn't go away, it grows with the app. For the scale of project this is, I can't commit to maintaining that debt long-term, and I don't want to merge something I can't sustain.

I did look at alternatives. Overlay services like Weglot/Localize.js, and even a "chrome-only" approach translating just navigation and common actions. But the overlay services don't fit the free tier (~5,300 words vs 2K limit) and add a monthly cost for something browsers already handle, and the chrome-only approach covers so little surface area that it's hard to justify even the small code change.

For now, built-in browser translation is what I'd recommend for anyone needing a different language. It requires zero code changes and works for any language out of the box.

I genuinely appreciate the time and care you put into this. If there are specific pages or workflows where browser translation falls down badly, I'd be open to hearing about that. There may be targeted, low-maintenance things we can do to help

In the future, I'd encourage opening an issue to discuss before taking on something this large, so we can align on approach and scope before the heavy lifting.

PR #419, opened 2026-05-17
`primaryMeterKey` for Quota Checkeres to control what gets shown on the provider page

IndianBoy42:
I noticed that when there is both a balance and allowance (eg crof, neuralwatt, apertis) then only the balance gets shown in the provider page. personally it made more sense to show the allowance for crof and neuralwatt (those get used first). I have no idea about apertis. the most flexible way that made sense to me was a primaryMeterKey attribute to control this


mcowger:
It's actually supposed to do both - if it's not it's a bug.  

This isn't the fix I'd want to do for that.

mcowger:
I've fixed this diffferently in #421 to show both when they are available.

PR #518, opened 2026-05-21
feat: add HTTP request headers and response status/headers to trace logs

### LINKED ISSUE #444: Perhaps I need to see the HTTP header information in the Trace logs.
When I use plexus to forward claude code requests,  
as shown in the figure, the upstream provide returns an exception, but plexus seems to treat it as a normal request, converting it into a normal end turn, leading to no further actions from Claude code.

I want to check the upstream http response header, but there is no way to view it. 

I plan to modify the Trace log structure so that not only can I see the http body, but also the header. 

Please take a look and see if it's feasible? @mcowger 

Image

micookie2:
Closes #444 

Enhance debug trace functionality to capture complete HTTP request and response metadata:

**Request logging:**
- Add requestHeaders parameter to startLog for capturing request headers
- Request structure: { headers: {...}, body: {...} }

**Response logging:**
- addReconstructedRawResponse wraps response as: { status, headers, body } (Only shown in "Raw Response (Reconstructed)" field)
- addRawResponse remains unchanged, only records body without metadata

**Changes:**
- debug-manager.ts: add addResponseMeta method, optimize response wrapping logic
- dispatcher.ts: call addResponseMeta at all response handling points
- sanitize-headers.ts: add sensitive header filtering utility
- Add comprehensive unit test coverage

### DIFF
--- a/packages/backend/drizzle/migrations/0046_add_debug_log_http_metadata.sql
+++ b/packages/backend/drizzle/migrations/0046_add_debug_log_http_metadata.sql
@@ -1,0 +1,3 @@
+ALTER TABLE `debug_logs` ADD `request_headers` text;--> statement-breakpoint
+ALTER TABLE `debug_logs` ADD `response_headers` text;--> statement-breakpoint
+ALTER TABLE `debug_logs` ADD `response_status` integer;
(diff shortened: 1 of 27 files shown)


micookie2:
Result：

image

mcowger (changes requested):
This is a good effort, thank you.  

Howeverm the intent of the 'recontructed' views is to have a valid JSON response.  Adding hte headers breaks that scenario.

I agree with storing this data, however it needs to be in a separate field.

PR #519, opened 2026-05-21
Fix OAuth per-provider stall overrides and failover config leak

sirn:
This PR fixes the following issues with stall detection:

## Per-provider stall configuration leak

Since `StallInspector` is shared between all providers in a request, when failover moves from Provider A (with per-provider overrides) to Provider B (without per-provider overrides), Provider A's overrides is leaked onto Provider B.

### Steps to reproduce

1. Make sure no stall configuration in global config
2. Configure Provider A (with a mock slow server) with a strict TTFB timeout/TTFB byte threshold, e.g. `5s/5000B`
3. Configure Provider B (with the same mock slow server) without TTFB timeout/TTFB byte threshold (use default)
4. Configure model alias with Provider A and B
5. Send a request to Provider A, fails, failover to Provider B
6. Provider B also treated as stall despite having no stall config (should use global)

### Root cause

The dispatcher has a guard around `addStallConfig`. When Provider B has no overrides, the `StallInspector` is never reconfigured. Additionally, when `resolveStallConfig` is `null` (stall detection off), `addStallConfig` skipped `updateConfig` instead of resetting it, so global "null" default (i.e. no stall check) is never applied. 

### Fix

Made sure `addStallConfig` is always called and make `addStallConfig` use the default stall settings when `resolveStallConfig` returns `null`.

## Per-provider stall configuration were not being applied to OAuth providers

OAuth providers have its own code path (`if (this.isPiAiRoute(...))`), but it was being called before provider stall overrides are computed. This means that OAuth providers will always use global stall configuration regardless of the presence of the stall configuration overrides.

### Steps to reproduce

1. Configure global stall configuration with a strict TTFB timeout/TTFB byte threshold, e.g. `5s/5000B`
2. Configure OAuth provider with a relaxed TTFB timeout/TTFB byte threshold, e.g. `30s/50B`
3. Send a streaming request to heavy OAuth models (e.g. `gpt-5.5`)
4. OAuth provider stalled despite having a very relaxed configuration

### Fix

Moved the per-provider wiring before dispatching to pi-ai. Also deduplicated `resolveStallConfig` logic and added a debug log.

## OAuth providers have no TTFB detection code path

`probeOAuthStreamStart` has no TTFB abort mechanism.

### Fix

Wired per-request TTFB abort controller inside `dispatchOAuthRequest`.

### DIFF
--- a/packages/backend/src/routes/inference/responses.ts
+++ b/packages/backend/src/routes/inference/responses.ts
@@ -178,7 +178,7 @@ export async function registerResponsesRoute(
         unifiedRequest,
         dispatchSignal,
         addTimeoutSource,
-        stallDetectionResult?.addStallConfig
+        stallDetectionResult.addStallConfig
       );
 
       // Emit 'updated' event with routing decision details
--- a/packages/backend/src/services/__tests__/oauth-probe-bookkeeping.test.ts
+++ b/packages/backend/src/services/__tests__/oauth-probe-bookkeeping.test.ts
@@ -36,8 +36,8 @@ async function consumeStream(stream: ReadableStream<any>): Promise<any[]> {
 /**
  * Helper to access the private probeOAuthStreamStart method for testing
  */
-function probeStreamStart(dispatcher: any, stream: ReadableStream<any>) {
-  return dispatcher.probeOAuthStreamStart(stream);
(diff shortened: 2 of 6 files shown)


mcowger (approved):
Nice catch

PR #583, opened 2026-06-12
feat(embeddings): add Gemini embeddings provider support

harrisony:
Adds a Gemini embeddings provider alongside the existing OpenAI pass-through. The embeddings pipeline now uses a transformer interface + factory (same pattern as chat/completions), so adding more provider types later is just implementing the interface.

## How it works

Clients still send standard OpenAI-format `POST /v1/embeddings`. When the router selects a Gemini provider (`api_type: gemini`), the `GeminiEmbeddingsTransformer`:

- Builds the endpoint: `/v1beta/models/{model}:embedContent` (single) or `:batchEmbedContents` (batch), prepending `models/` if needed
- Converts the request body to Gemini's `content: { parts: [{ text }] }` shape, with passthrough for `taskType`, `title`, and `outputDimensionality` when present
- Swaps `Authorization: Bearer` for `x-goog-api-key`
- Converts the response back to OpenAI format (`data[].embedding`, `usage.prompt_tokens`)

## What's changed

- `EmbeddingsTransformer` interface + `EmbeddingsTransformerFactory` — replaces the old monolithic `EmbeddingsTransformer` class
- `OpenAIEmbeddingsTransformer` — extracted pass-through behavior
- `GeminiEmbeddingsTransformer` — new, stateless (request threaded through `transformResponse` instead of stored as instance state)
- `dispatchEmbeddings` hardened: merges model `extraBody`, uses `parseJsonResponseBody`, calls `markProviderSuccess` for cooldown recovery, sets `isPassthrough: true`, and records `attemptCount`/`retryHistory`
- Router now includes `gemini`-type providers in embeddings matches; probe sends `input` field


mcowger (changes requested):
Why not use the Gemini API path?

/v1beta/models/<model name>:embedContent

PR #586, opened 2026-06-13
fix(quota): update Ollama checker for new settings page HTML structure

glade-at-gigwell:
The Ollama settings page changed its HTML structure. The old regex matching `style="width: X%"` immediately after label text no longer works because:

1. Progress bars are now deeper in nested divs with `data-usage-meter` attributes and trailing semicolons in style values
2. The percentage is now exposed via `aria-label` attributes like `aria-label="Session usage 4.4% used"` which is more stable
3. `data-time` reset timestamps are further from their labels (~2058+ chars)

**Changes:**
- **Primary extraction**: use `aria-label` regex (most robust against future HTML changes)
- **Fallback**: widen snippet window from 1000→2000 chars and relax style regex (no longer requires `%` immediately before closing `"`)
- **Reset time extraction**: widen search window from 1000→3000 chars

Verified against the current live Ollama settings page — extracts both Session (4.4%) and Weekly (60.1%) usage correctly, plus reset timestamps.

### DIFF
--- a/packages/backend/src/services/quota/checkers/ollama-checker.ts
+++ b/packages/backend/src/services/quota/checkers/ollama-checker.ts
@@ -3,14 +3,29 @@ import { z } from 'zod';
 import { logger } from '../../../utils/logger';
 
 function extractUsage(html: string, label: string): { percent: number; resetsAt?: string } | null {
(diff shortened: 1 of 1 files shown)


mcowger:
Nice.  Using the aria tags is a much better method.  Thanks!

PR #589, opened 2026-06-14
fix(mcp): immediately flush reponse headers

Amolith:
The MCP gateway proxied streaming responses with reply.send(stream). Fastify converts a web ReadableStream to a Node stream and forwards it through sendStream:

- https://github.com/fastify/fastify/blob/f15d4eaba03d86786bec2de7aed1236718c813f9/lib/reply.js#L684-L690

sendStream deliberately stages headers with res.setHeader() and defers res.writeHead() until there is data to send (so it can still emit an error status if the stream fails before producing output):

- https://github.com/fastify/fastify/blob/f15d4eaba03d86786bec2de7aed1236718c813f9/lib/reply.js#L733-L741

MCP streamable-HTTP servers open a long-lived GET SSE channel that is allowed to sit idle, sending no bytes until the server has a message to push. After writing the head, the Go SDK just blocks, writing nothing further until a message, cancellation, or close:

- https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#listening-for-messages-from-the-server
- https://github.com/modelcontextprotocol/go-sdk/blob/d3fd25b68a4b966b5f788a0b49e4ddf3e6048992/mcp/streamable.go#L959-L965

With an idle upstream, that first chunk never arrives, so Fastify never writes the response head and downstream clients (e.g. the MCP Go SDK client) block awaiting headers until they time out. Stateful servers, which depend on this GET channel, become unusable through the proxy while stateless ones work fine.

Soooo stream proxied responses through reply.raw with reply.hijack() + writeHead() + flushHeaders(), forwarding chunks directly so the head reaches the client immediately regardless of upstream stream timing, and cancelling the upstream reader on client disconnect so a departed client does not leave the upstream fetch connection (and, for stateful servers, its MCP session) open indefinitely.

I (da model) didn't know of a good way to test these changes other than a script using the same approach to communicate with the server I'm seeing the issue with, which it did and which confirmed the fix fwiw.

### DIFF
--- a/packages/backend/src/routes/mcp/index.ts
+++ b/packages/backend/src/routes/mcp/index.ts
@@ -14,12 +14,11 @@ const DEFAULT_TIMEOUT_MS = 120000;
 // reply.send() buffers a streamed response head until the first body chunk,
 // which strands clients on MCP's long-lived idle SSE channels.
 async function streamUpstreamResponse(
-  request: FastifyRequest,
   reply: FastifyReply,
   status: number,
   upstreamHeaders: Record<string, string>,
   stream: ReadableStream<Uint8Array>
-): Promise<FastifyReply> {
+): Promise<void> {
   const headers: Record<string, string> = { ...upstreamHeaders };
 
   // Preserve the upstream content-type (it carries the session-bound SSE
@@ -44,7 +43,7 @@ async function streamUpstreamResponse(
   const onClose = () => {
     reader.cancel().catch(() => {});
   };
-  request.raw.on('close', onClose);
+  reply.raw.on('close', onClose);
 
   try {
     for (;;) {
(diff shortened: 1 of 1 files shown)


mcowger (changes requested):
Superseded.

mcowger (changes requested):
The root cause analysis here is spot-on and `reply.hijack()` + `flushHeaders()` is the right fix. I ran all 43 MCP route tests against this change and they pass.

One issue that needs fixing before merge:

**`request.raw.on('close', ...)` is listening on the wrong lifecycle.** In HTTP/1.1, the incoming request's `close` event fires when request body consumption finishes — for a GET with no body, that's as soon as headers are parsed. This means the cleanup listener either fires immediately (canceling the upstream before any data is streamed) or has already fired before `.on()` is registered and never runs at all. For SSE, you want the response socket's close event instead: `reply.raw.on('close', ...)`.

Secondary: `return reply` at the bottom of `streamUpstreamResponse` is misleading since `hijack()` detaches the reply from Fastify's lifecycle — using `Promise<void>` would be clearer.

mcowger (approved):

PR #590, opened 2026-06-16
Fix/windows compile cli limit

edulml98:
Fix Windows binary compilation by replacing shell glob expansion with a small Windows-only helper. This avoids the Windows command-line length limit for migration assets while leaving the other OS compile commands unchanged.


mcowger:
@edulml98 - did you test and confirm this fix?

PR #633, opened 2026-06-25
fix(inference-v2): remap custom pi-ai provider ids for dispatch in 0.80

mcowger:
### **Unknown provider**
neuralwatt

___

## Problem

The upgrade to `@earendil-works/pi-ai` 0.80.1 (commit `26330d3`) introduced a regression for any Plexus provider configured with a custom `pi_ai_provider` id (e.g. `neuralwatt`, `wafer`).

**Before 0.80**, the executor called the `stream()`/`complete()` free functions from `@earendil-works/pi-ai/compat`, which dispatched by **`model.api`** (e.g. `openai-completions`), transparently routing custom providers through the appropriate builtin API handler.

**After 0.80**, the executor was changed to call `piAiModels.stream()` / `piAiModels.complete()` on a `builtinModels()` instance. This collection routes by **`model.provider`** via an internal map that only contains canonical builtin ids (`openai`, `anthropic`, `google`, …). Custom provider ids are not registered, so:

```
piAiModels.stream({ provider: "neuralwatt", api: "openai-completions", ... })
  → requireProvider("neuralwatt")
  → this.providers.get("neuralwatt") === undefined
  → throw new ModelsError("provider", "Unknown provider: neuralwatt")
```

Observed in production as request `c6f583c6-a61d-45bd-a07d-9c94fff973e7` (`glm-5.2` via `neuralwatt`).

## Fix

Add a `toDispatchModel()` helper in `pi-ai-utils.ts` that remaps `model.provider` to the canonical builtin that implements the same wire API before the model is handed to `piAiModels.stream()` / `.complete()`:

- If `model.provider` is already a registered builtin → return the model unchanged (same reference).
- Otherwise, look up `API_TO_BUILTIN_PROVIDER[model.api]` (map already existed for `resolvePiAiModel`) and return a shallow copy with the remapped provider.
- If the api has no mapping → return unchanged and let `piAiModels` surface its own error.

`pi-ai-executor.ts` now wraps both call sites with `toDispatchModel()`.

## Tests

- Added `getProvider` to `builtinModels` mock in `vitest.setup.ts` and the local mock in `resolve-pi-ai-model.test.ts`.
- Added a `toDispatchModel` describe block with 5 tests, including a regression case that mirrors the exact `neuralwatt`/`glm-5.2` failure (`openai-completions` → remapped to `openai`), plus coverage for `anthropic-messages`, `google-generative-ai`, builtin passthrough, and unknown-api fallthrough.

___

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -170,6 +170,11 @@ try {
   await configService.initialize();
   logger.debug('Configuration loaded from database');
 
+  // Register pi_ai_custom_providers with the piAiModels instance so custom
+  // providers can be dispatched directly by api type without remapping hacks.
+  const { registerCustomProvidersWithPiAi } = await import('./inference-v2/shared/pi-ai-utils');
+  await registerCustomProvidersWithPiAi();
+
   // One-time migration of legacy flat-format aliases to target groups.
   // TODO(#target-groups-cleanup): remove this after migration period.
   await configService.migrateLegacyTargetGroups();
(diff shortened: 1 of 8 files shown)


mcowger:
/pi what do you think of this PR?  No changes, just your thoughts.

PR #636, opened 2026-06-25
feat(context): Add opt-in context compaction for inference-v2 beta routes

darkspadez:
### **User description**
This PR adds request-side context compaction for /beta/ inference-v2 routes so oversized prompts can be reduced before provider dispatch. It introduces global, provider, and model/alias compaction settings with precedence-based resolution, plus native and headroom compaction strategies.

It also adds management API and frontend controls for compaction settings, emits compaction metadata headers on beta responses, and extends context-limit enforcement for the pi-ai path. The implementation is fail-open: compaction errors or non-reducing results fall back to the original context.

Tests cover settings resolution, repository/API round trips, native/headroom strategy behavior, context-limit enforcement, timeout/fail-open behavior, and beta route integration.

___

### **Description**
- Add opt-in context compaction for beta routes
  - Supports `native` and `headroom` compression strategies
  - Resolves settings hierarchically (alias > provider > global)

- Enforce context limits on inference-v2 path
  - Rejects over-window requests with 400 errors
  - Implements image- and tools-aware token estimation

- Introduce management API and frontend controls
  - Allows configuring compaction overrides in UI

- Add comprehensive test suite and documentation

___

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,9 +5,9 @@
     "": {
       "name": "plexus2",
       "devDependencies": {
-        "@biomejs/biome": "2.5.0",
-        "@redocly/cli": "^2.34.0",
-        "agent-browser": "^0.28.0",
+        "@biomejs/biome": "2.5.1",
+        "@redocly/cli": "^2.35.0",
+        "agent-browser": "^0.30.1",
         "bun-types": "1.3.14",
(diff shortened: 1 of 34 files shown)


darkspadez:
This will probably need more testing before merge.

PR #643, opened 2026-07-02
feat(inference-v2): add native vision fallthrough for beta pi-ai path

mcowger:
## Summary
Implements vision fallthrough for the beta `inference-v2` (pi-ai) path, operating directly on pi-ai's `Context`/`ImageContent`/`TextContent` types — no `UnifiedChatRequest` detour through the legacy `Dispatcher`.

## Changes
- **New:** `packages/backend/src/inference-v2/shared/vision-fallthrough.ts`
  - `contextHasImages(context)` — detects image blocks in a pi-ai `Context`.
  - `applyVisionFallthrough(context, descriptorModel, prompt, usageStorage?, parentMeta?)` — describes each image concurrently via a configured descriptor model, dispatched directly through `piAiModels.complete()` (candidate resolution → cooldown/concurrency gate → sequential failover across beta-compatible candidates). Returns a new `Context` with images replaced by `[Image Description: ...]` text blocks; never mutates the input.
  - LRU description cache keyed by `(model, sha256(imageData))`.
  - Records descriptor sub-call usage (`isDescriptorRequest: true`), mirroring the v1 `VisionDescriptorService`.
- **Modified:** `packages/backend/src/inference-v2/shared/pi-ai-executor.ts`
  - Wired vision fallthrough into the per-candidate failover loop, gated on the alias's `use_image_fallthrough` and the global `vision_fallthrough.descriptor_model` config — matching `dispatcher.ts`'s v1 semantics.
  - Runs before compaction/context-limit enforcement so token estimates reflect post-fallthrough (text-only) content.
  - Threaded `isVisionFallthrough`/`visionFallthroughModel` through all non-streaming and streaming (success/error/exception) usage records.

## Tests
- `vision-fallthrough.test.ts` — unit tests for `contextHasImages`/`applyVisionFallthrough` (multi-image ordering, caching, no-route error fallback).
- Extended `pi-ai-executor.test.ts` — executor-level tests confirming fallthrough triggers only for aliases with `use_image_fallthrough`, that the descriptor call happens before the target dispatch, and that images never reach a text-only target model.

Full backend suite (2091 tests), typecheck, and format checks all pass.

___

### **Description**
- Add native vision fallthrough module for inference-v2 pi-ai path

- Wire fallthrough into `pi-ai-executor` failover loop before compaction

- Thread `isVisionFallthrough` and `visionFallthroughModel` through usage records

- Add unit and executor-level integration tests for fallthrough behavior

___

### DIFF
--- a/packages/backend/src/inference-v2/shared/__tests__/pi-ai-executor.test.ts
+++ b/packages/backend/src/inference-v2/shared/__tests__/pi-ai-executor.test.ts
@@ -181,3 +181,215 @@ describe('runPiAiExecutor streaming errors', () => {
     );
   });
 });
+
+describe('runPiAiExecutor vision fallthrough', () => {
+  beforeEach(() => {
+    DebugManager.getInstance().resetForTesting();
+    setConfigForTesting({
+      providers: {
+        'text-only': {
+          api_base_url: 'https://api.openai.com/v1',
+          api_key: 'sk-test',
+          pi_ai_provider: 'openai-codex',
+          models: {
+            'text-model': {
+              pricing: { source: 'simple', input: 0, output: 0 },
+              pi_ai_model_id: 'text-model',
+            },
+          },
+        },
+        descriptor: {
+          api_base_url: 'https://api.openai.com/v1',
+          api_key: 'sk-test',
+          pi_ai_provider: 'openai-codex',
+          models: {
+            'gpt-4o': {
+              pricing: { source: 'simple', input: 0, output: 0 },
+              pi_ai_model_id: 'gpt-4o',
+            },
+          },
+        },
+      },
+      models: {
+        'text-model': {
+          priority: 'selector',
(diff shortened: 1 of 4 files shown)


mcowger:
Addressed both review points in b27ad51c:

- **Concurrency slot leak**: wrapped `buildPiAiModel` / `piAiModels.complete()` / usage-recording in a `doRelease()` guard (idempotent release flag), so a slot is released exactly once even if any of those calls throws synchronously — mirrors the pattern already used in `pi-ai-executor.ts`.
- **Cache pollution / repeated wasted calls**: now caching the `'No description available.'` placeholder when the descriptor model returns an empty response, since that's a deterministic model output for that image rather than a transient error. Genuine dispatch errors (network/API failures) are still left uncached so failover/retry can recover on the next request.

Added a regression test for the empty-description caching behavior. Full suite (2092 tests), typecheck, and format all pass.

PR #710, opened 2026-07-11
feat(dev): add production-like seed configuration

mcowger:
## Summary
Add a production-like default development configuration and tooling to regenerate or apply it. The fixtures retain realistic provider and model shapes while replacing provider credentials and session data with safe mock values.

## Changes
- **Seed data**: replace the existing local-focused defaults with realistic providers, quotas, aliases, and inference keys.
- **Populate tooling**: add a script that applies default fixtures plus optional user overrides to a running worktree-safe development instance.
- **Fixture generation**: add a staging-backup transformer that strips credentials, session tokens, and null values before writing committed defaults.
- **API typing**: add generated OpenAPI types used by the populate script and commands to regenerate them.
- **Documentation**: update contributor seed counts and clarify that committed credentials are redacted mocks.
- **Agent skills**: align frontend startup and management API examples with the canonical commands and staging environment variables.

## Notes
The fixture-generation script expects `/tmp/staging-config.json`. Applying the populated configuration requires a running Plexus development instance.


mcowger:
Addressed the review findings in `eef73d2c`:

- Replaced all committed inference-key secrets with deterministic dev-only values and made the fixture builder redact source key secrets.
- Rewrote provider and quota-checker endpoints to localhost mock URLs; quota checkers are disabled in committed defaults.
- Updated `PLEXUS_URL` handling so an existing port is preserved and `PLEXUS_PORT` is only applied when needed.
- Added regression coverage for secret redaction, local-only provider URLs, and URL/port resolution.

Live staging key rotation is intentionally deferred so OWUI and GitHub Actions credentials can be updated in coordination with the rotation.

mcowger:
Addressed the follow-up review in `5a618ffa`:

- Added recursive sensitive-field redaction, including arbitrary provider headers and nested token/secret/credential fields, while preserving non-sensitive fields.
- Added scheme-less hostname normalization for `PLEXUS_URL` and coverage for hostname, localhost, HTTPS, and pre-qualified-port cases.
- Changed populate execution to stop at each failed dependency boundary instead of creating aliases or keys against missing prerequisites.
- Expanded regression tests for nested secrets and custom headers.

Live staging keys remain unchanged as requested for coordinated rotation later.

mcowger:
Addressed the latest findings in `9921d349`:

- Structured sensitive fields now redact every primitive in the entire subtree while preserving object/array shape and primitive types.
- Added coverage for object-valued credentials and structured header values.
- Fixture generation now filters each key's `allowedModels` to seeded alias slugs and additional aliases.
- Removed the three unroutable OWUI model entries and added a committed-fixture routing invariant test.

Live staging keys remain unchanged as requested.

PR #739, opened 2026-07-17
fix(dispatch): aggregate Codex SSE for non-stream clients

uttkarsh-26:
## Summary

- recover Responses SSE returned to non-streaming clients, including when the upstream response is mislabeled as `application/json`
- rebuild an empty `response.completed.response.output` from ordered `response.output_item.done` events
- keep a non-empty completion payload authoritative and preserve the existing JSON/error paths

## Problem

The native Codex OAuth backend is stream-only. Plexus forces `stream: true` upstream, but the non-streaming dispatch path still attempts to parse the buffered response as JSON. In practice the backend can return SSE with `Content-Type: application/json`; additionally, the final `response.completed` event can have an empty `output` array while completed output items were emitted earlier. This produced HTTP 200 responses with null/empty assistant content for both `/v1/chat/completions` and `/v1/responses` clients using `stream: false`.

## Verification

- `bun run --cwd packages/backend test:force-all` — 215 files, 2209 tests passed
- `bun run typecheck` — passed for shared, frontend, and backend workspaces
- targeted native Codex dispatcher suite — 9 tests passed
- Biome formatting check for both changed files — passed
- live local Codex OAuth smoke tests:
  - non-streaming `/v1/chat/completions` returned the expected assistant text
  - non-streaming `/v1/responses` returned the expected output text

### DIFF
--- a/packages/backend/src/services/__tests__/dispatcher-codex-native.test.ts
+++ b/packages/backend/src/services/__tests__/dispatcher-codex-native.test.ts
@@ -30,6 +30,21 @@ const UPSTREAM_SSE = [
   '',
 ].join('\n');
 
+// The real Codex backend can leave response.completed.response.output empty;
+// the completed output item is carried by response.output_item.done instead.
+const UPSTREAM_SSE_EMPTY_COMPLETION_OUTPUT = [
+  'event: response.created',
+  'data: {"type":"response.created","response":{"id":"resp_abc","object":"response","status":"in_progress","model":"gpt-5-codex","output":[]}}',
+  '',
+  'event: response.output_item.done',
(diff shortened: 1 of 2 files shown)


uttkarsh-26:
@mcowger Review requested when you have a chance. This is a focused dispatcher fix with regression coverage; full verification details are in the PR body.

mcowger:
Excellent fix. The SSE recovery is narrowly scoped and robust: it handles the real-world `application/json` mislabeling, reconstructs missing completion output from ordered `response.output_item.done` events, and preserves authoritative completion payloads plus existing JSON/error paths. The focused regression coverage and verification are especially strong.

mcowger:
Released in 2026.07.19.2

PR #748, opened 2026-07-24
fix(gemini): signal malformed function calls

mcowger:
## Summary
Detect Gemini `MALFORMED_FUNCTION_CALL` terminal responses and signal them to clients as transient, retryable upstream errors. This prevents tool-heavy Gemini sessions from ending with a terminal unknown error while deliberately avoiding Plexus failover and provider cooldown behavior.

## Why / Context
Gemini 3.x can intermittently leak an internal tool-call representation as text and terminate its own candidate with `MALFORMED_FUNCTION_CALL`. Forwarding that finish reason verbatim is non-retryable in clients such as pi even though regenerating the turn usually succeeds.

## Changes
- **Detection**: treat Gemini's terminal `finishReason` as authoritative and record leaked `default_api` call syntax as secondary telemetry, including calls glued to preceding text.
- **Unary responses**: return protocol-appropriate HTTP 503 errors for Gemini, OpenAI, Anthropic, and Responses clients.
- **Streaming responses**: emit terminal protocol-specific retryable errors instead of normal completion events.
- **Dispatcher bypass**: inspect Gemini streams before the `bypassTransformation` branch, preserving ordinary stream bytes and rewriting only malformed terminal frames with a retryable message.
- **Routing health**: record the defect as a client-signaled error without triggering failover, cooldown, or provider-failure handling.
- **Safety**: do not reconstruct or execute leaked tool calls whose Python-like argument syntax cannot be parsed reliably.

## Testing
- Added regression coverage for authoritative detection, glued text leaks, unchanged ordinary bypass streams, malformed bypass terminal rewriting, unary 503 responses, and OpenAI/Anthropic streaming error formats.

### DIFF
--- a/packages/backend/src/services/__tests__/usage-logging.test.ts
+++ b/packages/backend/src/services/__tests__/usage-logging.test.ts
@@ -396,6 +396,37 @@ describe('UsageInspector', () => {
       expect(capturedRecord!.tokensInput).toBeGreaterThan(0);
       expect(capturedRecord!.tokensEstimated).toBe(1);
     });
+
+    it('does not update performance metrics for errored streams', async () => {
+      const requestId = 'test-error-performance-metrics';
+      const inspector = new UsageInspector(
+        requestId,
+        mockStorage,
+        {
+          requestId,
+          provider: 'google',
+          selectedModelName: 'gemini-3.6-flash',
+          responseStatus: 'error',
+        } as Partial<UsageRecord>,
+        mockPricing,
+        undefined,
+        Date.now() - 100,
+        false,
+        'gemini',
+        undefined,
+        undefined,
+        DEFAULT_GPU_PARAMS,
+        DEFAULT_MODEL
+      );
+
+      const source = new PassThrough();
+      source.pipe(inspector);
+      source.end();
+      await new Promise((resolve) => setTimeout(resolve, 100));
+
+      expect(mockStorage.saveRequest).toHaveBeenCalled();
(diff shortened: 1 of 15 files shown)


mcowger:
Addressed the quota-accounting concern in `558798f6`.

Unary malformed responses now run through usage finalization with `responseStatus: error`, preserving token/cost extraction and `recordQuotaUsage` while explicitly skipping successful performance metrics. The regression test verifies quota recording and error status. Quota headers were already populated before this branch and remain unchanged.

mcowger:
Addressed the updated review in `96a2c529`:

- `UsageInspector` now updates provider performance metrics only when `responseStatus === "success"`, so client-signaled malformed streams remain errors and do not influence successful provider performance. Added a focused regression.
- Ordinary CRLF bypass bytes were already preserved because splitting on `\n` retains the preceding `\r`. Added an exact CRLF byte-preservation regression to make that behavior explicit; no rewriter change was necessary.

mcowger:
The remaining “Stream status reset” finding is not applicable to the streaming path:

- `finalizeUsage()` (including `usageRecord.responseStatus = options.responseStatus ?? "success"`) is called only by the non-streaming branch.
- Streaming responses are finalized by `UsageInspector._flush()`, which does not overwrite `responseStatus`.
- The malformed-stream callback mutates the same `usageRecord` held by `UsageInspector`, setting it to `error` before `_flush()`.
- `96a2c529` additionally gates stream performance updates on `responseStatus === "success"`, with a regression covering the error case.

No further code change is needed for this finding.

PR #758, opened 2026-07-27
fix(debug): persist ephemeral traces when a debug dimension matches

TheArchitectit:
## Problem

Requests whose provider doesn't return token usage need token estimation. Those requests are marked **ephemeral** (`response-handler.ts`) so their debug trace is captured in memory for estimation and never stored.

The bug: the ephemeral marker suppressed persistence **even when an operator-enabled debug dimension matched the request** — global debug mode, a targeted key/alias/provider, or a capture-on-error `forcePersist`. Traces the operator explicitly asked for were silently dropped:

- `DebugManager.flush()` skipped ephemeral requests unconditionally, before checking any enable dimension
- `discardEphemeral()` (called by the usage inspector right after estimation) deleted the pending trace outright

Verified against a production database (24h window):

```
tokens_estimated = 0:  4676 usage rows → 3888 have debug logs
tokens_estimated = 1:  1501 usage rows →    0 have debug logs
```

Every request needing token estimation was missing from the debug_logs view, and in capture-on-error mode even *error* traces were lost when the request also needed estimation.

## Fix

The persistence decision for both `flush()` and `discardEphemeral()` now goes through one shared predicate, `shouldPersistLog()`:

- any debug dimension matches (global / key / alias / provider), **or**
- the request was flagged for forced persistence (capture-on-error)

The ephemeral marker only wins when nothing wants the trace persisted. Behavior is unchanged when debug capture is fully off (ephemeral traces are still dropped, payloads still aren't persisted).

As a side effect, both paths now also remove the request id from `ephemeralRequests` on the persist and not-enabled paths, so the set doesn't accumulate ids.

## Tests

New suite `debug-manager-ephemeral.test.ts` (7 tests) covering: discard when nothing is enabled, persist under global capture, persist under key targeting, the usage-inspector ordering (`discardEphemeral` → `flush`), and both capture-on-error cases (with and without `forcePersist`).

Related existing suites pass: debug-manager-targets/startup/metadata, debug-logging-error/reconstruction, usage-logging(+metadata), speech-debug, transcriptions-debug — 44/44.

### DIFF
--- a/packages/backend/src/services/__tests__/debug-manager-ephemeral.test.ts
+++ b/packages/backend/src/services/__tests__/debug-manager-ephemeral.test.ts
@@ -1,0 +1,113 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest';
+import { DebugManager } from '../observability/debug-manager';
+import { runInRequestContext } from '../observability/request-context';
+import type { UsageStorageService } from '../observability/usage-storage';
+
+describe('DebugManager ephemeral capture', () => {
+  let debugManager: DebugManager;
+  let saveDebugLog: ReturnType<typeof vi.fn>;
+
+  beforeEach(() => {
+    debugManager = DebugManager.getInstance();
+    debugManager.resetForTesting();
+    debugManager.setEnabled(false);
+    saveDebugLog = vi.fn();
+    debugManager.setStorage({ saveDebugLog } as unknown as UsageStorageService);
+  });
+
+  test('discards the ephemeral trace when no debug dimension is enabled', () => {
+    debugManager.startLog('req-none', { model: 'untracked-alias' });
+    debugManager.markEphemeral('req-none');
+    debugManager.flush('req-none');
+
+    expect(saveDebugLog).not.toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


mcowger:
Pushed an additional commit (`30b32894`) fixing a related issue found during review:

`response-handler.ts` temporarily flips the **global** debug flag (`debugManager.setEnabled(true)`) for any request needing token estimation, then restores it — but that restore only existed on the client-error path and the streaming (`pipeline.on('error'/'end')`) path. The non-streaming (unary) success branch never restored it.

Before this PR that gap was low-impact, since `flush()` unconditionally skipped ephemeral traces regardless of the global flag. After this PR's fix, `shouldPersistLog()` checks `enabledGlobal` — so the first unary request needing token estimation (with debug otherwise off) would leave global debug mode stuck `true` forever (it's a process-wide singleton), causing **every** subsequent request's debug trace to get persisted, not just the ones matching an operator-enabled dimension.

Added the missing restore in the unary branch (mirroring the existing streaming-path restore) plus a regression test (`response-handler.test.ts`) asserting `debugManager.isEnabled()` returns to its prior state after a unary response with `shouldEstimateTokens=true`. Full backend suite (525 tests) passes.

PR #759, opened 2026-07-27
fix(responses): preserve streamed tool argument fragments

Zeng-Zer:
## Summary

Preserve streamed tool argument fragments when one chunk is independently valid JSON. This prevents Responses API tool calls from losing their accumulated prefix before finalization.

## Problem

`ResponsesTransformer.formatStream()` supports both incremental argument deltas and providers that repeat a complete argument snapshot. Its snapshot heuristic currently replaces the accumulator whenever one chunk parses as a complete JSON object.

Stream boundaries are arbitrary. A shell command or patch can contain an empty object literal that arrives as its own chunk. Plexus then mistakes that content fragment for a complete snapshot, discards every preceding argument fragment, and emits a truncated suffix in `response.output_item.done` and `response.completed`.

Clients receive malformed finalized arguments even though the preceding `response.function_call_arguments.delta` events reconstruct correctly. Defensive clients may reduce the malformed value to an empty object, causing required tool parameters such as `command` or `input` to disappear.

## Simplified example

Suppose a model streams one tool argument in these chunks:

```text
1. {"command":"before 
2. {}
3.  after"}
```

Concatenating them should produce:

```json
{"command":"before {} after"}
```

The previous heuristic parsed chunk 2 independently as valid JSON and replaced the accumulated prefix with it. Plexus then finalized this malformed suffix:

```text
{} after"}
```

A client could fail to parse that value and fall back to empty arguments, so required fields such as `command` or `input` disappeared even though the streamed preview looked complete.

## Changes

- Replace accumulated arguments only when a valid complete JSON candidate extends the accumulated prefix.
- Ignore leading whitespace when comparing a partial prefix with a complete snapshot.
- Add regression coverage for object-looking content inside a streamed string argument.
- Preserve coverage for complete snapshots following whitespace-prefixed partial data.

## Testing

- `bun run format:check`
- `bun run typecheck`
- `bun run test` (51 test files, 348 tests passed)

### DIFF
--- a/packages/backend/src/transformers/__tests__/responses-codex-tools.test.ts
+++ b/packages/backend/src/transformers/__tests__/responses-codex-tools.test.ts
@@ -37,6 +37,85 @@ async function collectFormatStreamEvents(
 }
 
 describe('Codex CLI namespace tool flattening', () => {
+  test('preserves JSON-looking text inside streamed tool arguments', async () => {
+    const transformer = new ResponsesTransformer();
+    const events = await collectFormatStreamEvents(transformer, [
+      {
+        id: 'resp_1',
+        model: 'gpt-5.6',
+        created: 1,
+        delta: {
+          tool_calls: [{ index: 0, id: 'call_1', function: { name: 'run', arguments: '' } }],
+        },
+        finish_reason: null,
+      },
+      {
+        id: 'resp_1',
+        model: 'gpt-5.6',
+        created: 1,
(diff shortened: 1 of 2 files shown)


mcowger:
Merged — thanks for the sharp catch and the clean fix here! The regression tests make the edge case (content fragments that look like complete JSON) really clear, and the prefix-extension check is a nice minimal way to close the gap without disturbing the legitimate snapshot-provider path. Appreciate the thorough writeup and testing notes too.

PR #767, opened 2026-07-28
fix(dispatch): sanitize foreign tool ids on Anthropic-bound requests

darkspadez:
## Summary

Sessions that switch models mid-conversation (subagents, review passes) accumulate tool-call IDs minted by non-Anthropic providers in the client's history — Moonshot/Kimi emits `functions.WebSearch:3`, pi-ai replays composite `call_x|fc_y` ids, Gemini uses dotted tool names as ids. Anthropic hard-400s any `tool_use.id` / `tool_result.tool_use_id` outside `^[a-zA-Z0-9_-]+$`, and because the same poisoned body goes to every failover target, the whole alias group is exhausted and the session is permanently stuck:

```
All targets failed: cc/claude-opus-5 (400), ... Last error: messages.222.content.1.tool_use.id: String should match pattern '^[a-zA-Z0-9_-]+$'
```

(Observed in production 2026-07-28: 29 of 100 recent gateway errors, one Claude Code session unable to continue at all.)

## What changed

- New provider adapter **`normalize_anthropic_tool_ids`** (`transformers/adapters/normalize-anthropic-tool-ids.adapter.ts`), auto-injected for every route whose **final** outbound wire type is Anthropic Messages (`getApiBaseType(effectiveApiType) === 'messages'`).
- `resolveAdapters(route, effectiveApiType?)` gains an optional second parameter; the sole production call site (`request-manager.ts`) passes the already-computed `effectiveApiType`. Existing single-arg callers keep compiling and get no format-scoped injection.
- Running in the `preDispatch` loop means one implementation covers all four Anthropic-bound dispatch shapes — **messages↔messages pass-through** (the failing prod path, where the anthropic request-builder never runs), cross-format transforms, native OAuth, and masked API-key routes — and lands **before** `prepareNativeOAuthDispatch`, so the CC masking billing signature is computed over the sanitized body.

### Sanitizer semantics

Pure, deterministic, idempotent. Charset-valid ids of any length pass through untouched (Anthropic's own `toolu_`/`srvtoolu_` ids are never rewritten). Offenders get invalid chars replaced with `_` plus an 8-hex sha256 suffix of the original (utf16le-encoded so lone surrogates stay distinct), capped at 64 chars, e.g. `functions.WebSearch:3` → `functions_WebSearch_3_0ad6c974`. Purity keeps `tool_use`/`tool_result` pairs consistent with zero state — across turns, same-target strip-retries, and failover attempts. Values that refuse string coercion (`{"toString": null}` is valid JSON) fall back to a deterministic marker instead of throwing. Proactive-only by the same decision rule as the GPT-5 option suppression NOTE: the defect is visible in the request and rejected identically by every messages-format target, and a reactive post-400 mutation would run after CCH signing.

## Testing

- 17 unit tests on the sanitizer/walker (charset edge cases, lone-surrogate distinctness, hostile coercion, idempotence, pair consistency, chat/responses bodies untouched).
- 7 resolver-injection tests (base type + subtype scoping, disable/re-enable mechanism, implicit ordering); all 21 pre-existing resolver tests pass unmodified.
- 6 wire-level dispatcher tests against the real `Dispatcher` with mocked fetch: pass-through, masked route (including recomputing the CC billing hash over the transmitted body to prove sign-after-sanitize ordering, with an anti-vacuity control), chat→anthropic transform, non-anthropic control (byte-identical), and failover re-sanitization per attempt.
- Full suite: 235 files / 2608 tests green; typecheck and biome clean.

### DIFF
--- a/packages/backend/src/services/dispatch/adapter-resolver.ts
+++ b/packages/backend/src/services/dispatch/adapter-resolver.ts
@@ -1,9 +1,11 @@
-import type { ProviderAdapter, ResolvedAdapter } from '../../types/provider-adapter';
+import type { ResolvedAdapter } from '../../types/provider-adapter';
 import type { RouteResult } from '../routing/router';
 import type { AdapterEntry } from '../../config';
 import { ADAPTER_REGISTRY } from '../../transformers/adapters/index';
+import { normalizeAnthropicToolIdsAdapter } from '../../transformers/adapters/normalize-anthropic-tool-ids.adapter';
 import { stripUnsupportedToolSearchAdapter } from '../../transformers/adapters/strip-unsupported-tool-search.adapter';
 import { suppressUnsupportedGpt5OptionsAdapter } from '../../transformers/adapters/suppress-unsupported-gpt5-options.adapter';
+import { getApiBaseType } from '../../utils/api-format';
 import { logger } from '../../utils/logger';
 
 /**
@@ -12,8 +14,10 @@ import { logger } from '../../utils/logger';
  * Resolution order:
  *   1. Implicit adapters automatically injected for the route's target
  *      provider (currently: tool-search stripping for `pi_ai_provider ===
- *      'openrouter'`). These run first so user-configured adapters see the
- *      cleaned-up payload if they inspect it.
+ *      'openrouter'`), its model (GPT-5 option suppression) and its outbound
+ *      wire format (Anthropic tool-id normalization for every route whose
+ *      final wire type is Anthropic Messages). These run first so
+ *      user-configured adapters see the cleaned-up payload if they inspect it.
  *   2. Provider-level `adapter` (applies to all models under the provider)
  *   3. Model-level `adapter`   (appended after provider-level adapters)
  *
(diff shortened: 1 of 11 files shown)


mcowger (dismissed):
Right solution, wrong scope. Not all messages targets are anthropic, and thus should not globally receive this change.  

Please adjust:

1. Add UI to enable / disable per provider
2. If you like, enable by default if provider base url contains anthropic.com

packages/backend/src/services/dispatch/adapter-resolver.ts
```diff
@@ -57,30 +73,144 @@ export function resolveAdapters(route: RouteResult): ResolvedAdapter[] {
...
+  effectiveApiType?: string
+): string[] {
+  if (typeof base === 'string') return [base];
+  const urlMap = base ?? {};
+  if (!effectiveApiType) {
+    return Object.values(urlMap).filter((value): value is string => typeof value === 'string');
+  }
+  const typeKey = effectiveApiType.toLowerCase();
+  const selected = urlMap[typeKey] || urlMap[getApiBaseType(typeKey)];
+  return typeof selected === 'string' && selected.length > 0 ? [selected] : [];
```
mcowger, line 211: **P1 — Use the actual dispatch URL for this gate.** `selectDispatchUrls` intentionally stops after exact/base-type lookup, but `resolveProviderBaseUrl` continues through `default`, aliases, and first-key fallback. That means a valid Messages route such as a model with `access_via: ['messages']` and `api_base_url: { default: 'https://api.anthropic.com' }` is dispatched to Anthropic while this returns `[]`; foreign tool IDs remain unsanitized and reproduce the hard 400 this PR fixes. Mixed-case map keys have the same issue because dispatch falls back to the first key. Please share the dispatch URL-resolution logic (or otherwise exactly mirror it) before testing whether the selected URL is Anthropic.

mcowger (dismissed):
The two-part `messages && Anthropic-like target` condition applies only to **automatic** injection; the provider UI’s **Enabled** mode force-enables the adapter for a non-`anthropic.com` Messages gateway, so that part matches the requested scope. I found one blocking gap in automatic detection: it does not use the same URL fallback rules as dispatch, allowing some requests actually sent to Anthropic to bypass normalization.

PR #787, opened 2026-08-03
feat(mcp-oauth): add OAuth 2.1 authorization-code flow for MCP proxy

gwyntel-git:
## Summary
Adds an optional, standard **OAuth 2.1 authorization-code flow** for the MCP proxy endpoints. When enabled, MCP clients authenticate through an OAuth provider instead of (or in addition to) a plain API key. The flow follows the Model Context Protocol's OAuth authorization specification (RFC 8707 resource indicators, PKCE, RFC 9728 token exchange conventions) and is fully compatible with an **OAuth Authorization Server Metadata** / `/.well-known/oauth-protected-resource` discovery document.

The feature is **opt-in** (`mcpOAuth.enabled`, default `false`). With it off, Plexus behaves exactly as it does today. With it on, Plexus publishes an OAuth-protected resource, issues clients, and validates access tokens against its own IdP.

## Why / Context
MCP is an emerging standard for letting LLM clients reach external tools and data. The ecosystem is converging on **OAuth-based authentication** for MCP servers — clients like Claude, Cursor, and other model-native tools expect to negotiate auth via an authorization server, not by pasting a static key into a config file.

Today Plexus only supports API-key auth for its MCP endpoints. That works for a single trusted user but does not fit a hosted gateway that wants to expose MCP servers to many independent clients: there is no way to issue per-client credentials, revoke a single client, rotate keys, or follow the standard discovery flow. This change lets a Plexus instance act as both the MCP server and an OAuth provider (via a pluggable `AuthProvider` interface), so any standards-compliant MCP client can connect through the normal OAuth dance — including first-party clients that need scoped, revocable, expiring tokens.

## Changes
- **Pluggable auth layer** (`services/mcp-oauth/`): an `AuthProvider` interface, a factory that resolves the configured provider, and a bundled `plexus-idp-provider` (Plexus's own OAuth IdP) implementation.
- **MCP route integration** (`routes/mcp/index.ts`): when OAuth is enabled, the MCP endpoints run the authorization-code flow and require a valid access token; otherwise they fall back to the existing API-key auth path unchanged.
- **IdP storage** (`db/mcp-oauth-repository.ts` + schema): `mcp_oauth_clients`, `mcp_oauth_tokens`, and `mcp_oauth_authorization_codes` tables (SQLite and Postgres). Tokens store a hash of the bound API key secret (`apiKeySecretHash`) so rotation can be detected, and `key_name` is indexed.
- **Key-rotation / revocation**: rotating or deleting an API key eagerly revokes every access/refresh token bound to it (`revokedAt` set immediately), so a leaked key cannot leave live OAuth sessions behind.
- **Management API** (`routes/management/mcp-oauth.ts`): endpoints to create/list OAuth clients, and to view/revoke issued tokens from the Admin UI.
- **Discovery**: when enabled, Plexus serves the OAuth discovery metadata (`/.well-known/oauth-protected-resource`, `/.well-known/openid-configuration`) describing the protected resource and the authorization server.
- **Config** (`config.ts`): new `mcpOAuth` block (`enabled`, issuer, provider selection, and related settings).
- **Frontend** (`pages/Config.tsx`, `pages/Mcp.tsx`, `lib/api.ts`): admin UI to toggle MCP OAuth, manage clients, and inspect/revoke tokens.

## Testing
- New unit tests cover the `plexus-idp-provider` (token validation, client creation) and the DB-backed key-revocation behavior (tokens are revoked the moment a bound API key rotates or is deleted).
- MCP route tests cover both auth paths: API-key fallback (feature off) and OAuth enforcement (feature on).
- Full backend suite passes: **243 test files, 2,680 tests green**, including the 33 MCP OAuth tests. Typecheck and frontend build pass via the pre-commit hooks.

## Notes / Follow-ups
- **Schema only, no committed migrations.** Per `CONTRIBUTING.md`, migrations are auto-generated by CI after merge — this PR touches only the Drizzle schema `.ts` files.
- Default-off keeps existing deployments byte-for-byte behaviorally unchanged.
- The disabled path changes slightly: the static discovery stubs (`/.well-known/*`) previously returned canned data; with OAuth off those endpoints now return 404. Flagging for reviewer awareness.
- Provider resolution is behind an interface so a future provider (e.g. an external OAuth server) can be added without touching the route layer.

---

## Update: Review fixes + client lifecycle UI (0a710491)

All 8 review comments addressed in 3bb390d6, plus:

**Client lifecycle Admin UI** (0a710491):
- Status badge (Active/Disabled) on each OAuth client card
- Disable/enable toggle, revoke-all-tokens, and delete client buttons
- Disabled clients dimmed for visual clarity
- All destructive actions behind confirmation dialogs

**Comprehensive test coverage** (0a710491):
- Unit tests for all 8 new behaviors (expired/disabled key rejection, scope constraint, resource validation, disabled client rejection, code invalidation on rotation, cache headers)
- Route tests for scope enforcement at the protected resource and discovery endpoint consistency
- 243 test files / 2,695 tests passing

### DIFF
--- a/packages/backend/src/routes/management/mcp-oauth.ts
+++ b/packages/backend/src/routes/management/mcp-oauth.ts
@@ -1,0 +1,37 @@
+import { FastifyInstance } from 'fastify';
+import { z } from 'zod';
+import { McpOauthRepository } from '../../db/mcp-oauth-repository';
+
+const revokeTokenParamsSchema = z.object({
+  id: z.coerce.number().int().positive(),
+});
+
+export async function registerMcpOAuthManagementRoutes(fastify: FastifyInstance) {
+  const repo = new McpOauthRepository();
+
+  fastify.get('/v0/management/mcp-oauth/clients', async (_request, reply) => {
+    try {
+      const clients = await repo.listClientsWithActiveTokens();
+      return reply.send({ clients });
+    } catch (e: any) {
+      return reply.code(500).send({ error: e.message || 'Internal server error' });
+    }
+  });
+
+  fastify.post('/v0/management/mcp-oauth/tokens/:id/revoke', async (request, reply) => {
+    const parsed = revokeTokenParamsSchema.safeParse(request.params);
+    if (!parsed.success) {
+      return reply.code(400).send({ error: 'Invalid token id', details: parsed.error.issues });
+    }
+
+    try {
+      const revokedCount = await repo.revokeTokenById(parsed.data.id);
+      if (revokedCount === 0) {
+        return reply.code(404).send({ error: 'Active OAuth token not found' });
+      }
+      return reply.send({ success: true, revokedCount });
+    } catch (e: any) {
+      return reply.code(500).send({ error: e.message || 'Internal server error' });
+    }
+  });
+}
--- a/packages/backend/src/routes/management/mcp-oauth.ts
+++ b/packages/backend/src/routes/management/mcp-oauth.ts
@@ -6,6 +6,14 @@ const revokeTokenParamsSchema = z.object({
   id: z.coerce.number().int().positive(),
 });
 
+const clientParamsSchema = z.object({
+  clientId: z.string().min(1),
+});
+
+const updateClientStatusSchema = z.object({
+  status: z.enum(['active', 'disabled']),
+});
+
 export async function registerMcpOAuthManagementRoutes(fastify: FastifyInstance) {
   const repo = new McpOauthRepository();
 
@@ -13,6 +21,70 @@ export async function registerMcpOAuthManagementRoutes(fastify: FastifyInstance)
     try {
       const clients = await repo.listClientsWithActiveTokens();
       return reply.send({ clients });
+    } catch (e: any) {
+      return reply.code(500).send({ error: e.message || 'Internal server error' });
+    }
+  });
+
+  fastify.patch('/v0/management/mcp-oauth/clients/:clientId', async (request, reply) => {
+    const parsedParams = clientParamsSchema.safeParse(request.params);
+    if (!parsedParams.success) {
+      return reply.code(400).send({ error: 'Invalid client id' });
+    }
+    const parsedBody = updateClientStatusSchema.safeParse(request.body ?? {});
+    if (!parsedBody.success) {
+      return reply.code(400).send({ error: 'Invalid status', details: parsedBody.error.issues });
+    }
+
+    const { clientId } = parsedParams.data;
+    const { status } = parsedBody.data;
+
+    try {
+      const client = await repo.getClient(clientId);
+      if (!client) {
+        return reply.code(404).send({ error: 'OAuth client not found' });
+      }
+      await repo.setClientStatus(clientId, status);
+      return reply.send({ success: true, clientId, status });
+    } catch (e: any) {
+      return reply.code(500).send({ error: e.message || 'Internal server error' });
+    }
+  });
+
+  fastify.delete('/v0/management/mcp-oauth/clients/:clientId', async (request, reply) => {
+    const parsed = clientParamsSchema.safeParse(request.params);
+    if (!parsed.success) {
+      return reply.code(400).send({ error: 'Invalid client id' });
+    }
+
+    const { clientId } = parsed.data;
+    try {
+      const client = await repo.getClient(clientId);
+      if (!client) {
+        return reply.code(404).send({ error: 'OAuth client not found' });
+      }
+      const revokedCount = await repo.revokeTokensForClient(clientId);
+      await repo.deleteClient(clientId);
+      return reply.send({ success: true, clientId, revokedCount });
+    } catch (e: any) {
+      return reply.code(500).send({ error: e.message || 'Internal server error' });
+    }
+  });
+
+  fastify.post('/v0/management/mcp-oauth/clients/:clientId/revoke', async (request, reply) => {
+    const parsed = clientParamsSchema.safeParse(request.params);
+    if (!parsed.success) {
+      return reply.code(400).send({ error: 'Invalid client id' });
+    }
+
+    const { clientId } = parsed.data;
+    try {
+      const client = await repo.getClient(clientId);
+      if (!client) {
+        return reply.code(404).send({ error: 'OAuth client not found' });
+      }
+      const revokedCount = await repo.revokeTokensForClient(clientId);
+      return reply.send({ success: true, clientId, revokedCount });
     } catch (e: any) {
       return reply.code(500).send({ error: e.message || 'Internal server error' });
     }
--- a/packages/backend/src/routes/mcp/index.ts
+++ b/packages/backend/src/routes/mcp/index.ts
@@ -1,13 +1,103 @@
 import { FastifyInstance, FastifyRequest, FastifyReply } from 'fastify';
 import bearerAuth from '@fastify/bearer-auth';
-import { createAuthHook } from '../../utils/auth';
+import formbody from '@fastify/formbody';
+import { getConfig } from '../../config';
+import {
+  attachPlexusApiKeyAuth,
+  createAuthHook,
+  isRequestIpAllowed,
+  validatePlexusApiKey,
+} from '../../utils/auth';
 import { logger } from '../../utils/logger';
 import * as mcpProxyService from '../../services/mcp-proxy/mcp-proxy-service';
 import { getClientIp } from '../../utils/ip';
 import { McpUsageStorageService } from '../../services/mcp-proxy/mcp-usage-storage';
 import { registerPlexusMcpRoutes } from './plexus';
+import { getMcpAuthProvider, isMcpOAuthEnabled } from '../../services/mcp-oauth/provider-factory';
+import { getMcpResourceUrl, getRequestBaseUrl } from '../../services/mcp-oauth/url';
+import { MCP_OAUTH_ACCESS_TOKEN_PREFIX } from '../../services/mcp-oauth/plexus-idp-provider';
 
 const DEFAULT_TIMEOUT_MS = 120000;
+
+function authErrorResponse(message: string) {
+  return { error: { message, type: 'auth_error', code: 401 } };
+}
+
+function getStringHeader(value: string | string[] | undefined): string | null {
+  if (Array.isArray(value)) return value[0] ?? null;
+  return value ?? null;
+}
+
+function extractBearerCredential(authorization: string): string {
+  return authorization.toLowerCase().startsWith('bearer ')
+    ? authorization.slice('bearer '.length)
+    : authorization;
+}
+
+function setInitialOAuthChallenge(request: FastifyRequest, reply: FastifyReply) {
+  const metadataUrl = `${getRequestBaseUrl(request)}/.well-known/oauth-protected-resource`;
+  reply.header('WWW-Authenticate', `Bearer resource_metadata="${metadataUrl}"`);
+}
+
+function setInvalidTokenChallenge(reply: FastifyReply) {
+  reply.header('WWW-Authenticate', 'Bearer error="invalid_token"');
+}
+
+async function mcpOAuthFallbackAuth(request: FastifyRequest, reply: FastifyReply) {
+  const authorization = getStringHeader(request.headers.authorization);
+  const xApiKey = getStringHeader(request.headers['x-api-key']);
+  const xGoogApiKey = getStringHeader(request.headers['x-goog-api-key']);
+  const queryKey =
+    request.query && typeof request.query === 'object'
+      ? typeof (request.query as any).key === 'string'
+        ? (request.query as any).key
+        : null
+      : null;
+
+  const tryRawApiKey = (secret: string): boolean => {
+    const result = validatePlexusApiKey(secret, request);
+    if (!result) return false;
+    attachPlexusApiKeyAuth(request, result);
+    return true;
+  };
+
+  if (authorization) {
+    const credential = extractBearerCredential(authorization);
+    if (tryRawApiKey(credential)) return;
+
+    const provider = getMcpAuthProvider();
+    const oauthResult = credential.startsWith(MCP_OAUTH_ACCESS_TOKEN_PREFIX)
+      ? await provider?.validateToken(credential)
+      : null;
+    if (oauthResult) {
+      const config = getConfig();
+      const keyConfig = config.keys?.[oauthResult.keyName];
+      if (keyConfig && isRequestIpAllowed(request, keyConfig.allowedIps, config.trustedProxies)) {
+        attachPlexusApiKeyAuth(request, {
+          keyName: oauthResult.keyName,
+          keyConfig,
+          attribution: null,
+        });
+        return;
+      }
+    }
+
+    setInvalidTokenChallenge(reply);
+    await reply.code(401).send(authErrorResponse('Invalid bearer token'));
+    return reply;
+  }
+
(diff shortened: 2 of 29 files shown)


gwyntel-git:
Relates to issue #788 — this PR implements the MCP OAuth 2.1 authorization-code flow described there.

gwyntel-git:
Fixed the appearance — cloned the MCP OAuth consent page onto the Plexus design system (real PlexusMark, glass card, amber gradient wordmark/button, matching design tokens and font stacks).

packages/backend/src/routes/mcp/index.ts
```diff
@@ -67,61 +157,48 @@ export async function registerMcpRoutes(
...
-    return reply.send({
-      issuer: '/',
-      authorization_endpoint: '/oauth/authorize',
-      token_endpoint: '/oauth/token',
-      grant_types_supported: ['client_credentials', 'bearer'],
-      token_endpoint_auth_methods_supported: ['client_secret_basic', 'client_secret_post', 'none'],
-      resource_supported: true,
-    });
-  });
+  const oauthEnabled = isMcpOAuthEnabled();
```
mcowger: [P1] Make the OAuth setting effective without a restart. `oauthEnabled` and `authProvider` are captured once while Fastify registers the routes. The Config UI can persist a changed `mcpOAuth.enabled` value, but that change cannot add or remove the OAuth endpoints or switch the MCP authentication hook until the process restarts. Please resolve this setting at request time, or make it explicitly restart-only and communicate that in the UI.

packages/backend/src/services/mcp-oauth/plexus-idp-provider.ts
```diff
@@ -0,0 +1,573 @@
...
+  }
+
+  async validateToken(token: string): Promise<{ keyName: string; scopes: string[] } | null> {
+    if (!token.startsWith(ACCESS_TOKEN_PREFIX)) return null;
+
+    const record = await this.repo.getAccessToken(token);
+    if (!record) return null;
+    if (record.revokedAt !== null) return null;
+    if (record.accessTokenExpiresAt <= Date.now()) return null;
+    if (!this.isTokenBoundToCurrentApiKeySecret(record.keyName, record.apiKeySecretHash))
```
mcowger, line 371: [P1] Reject tokens for expired or disabled API keys. Token validation checks token expiry, revocation, and secret-hash binding, but it does not check whether the underlying API key is expired or disabled. The MCP route then attaches the key principal after only checking the IP allowlist. Consequently, an OAuth access token remains usable after its API key expires or is disabled, unless the secret is also rotated or deleted. Please apply the same key-lifecycle checks used by `validatePlexusApiKey`.

packages/backend/src/routes/mcp/index.ts
```diff
@@ -1,14 +1,104 @@
...
+
+    const provider = getMcpAuthProvider();
+    const oauthResult = credential.startsWith(MCP_OAUTH_ACCESS_TOKEN_PREFIX)
+      ? await provider?.validateToken(credential)
+      : null;
+    if (oauthResult) {
+      const config = getConfig();
+      const keyConfig = config.keys?.[oauthResult.keyName];
+      if (keyConfig && isRequestIpAllowed(request, keyConfig.allowedIps, config.trustedProxies)) {
+        attachPlexusApiKeyAuth(request, {
```
mcowger, line 98: [P1] Enforce OAuth scopes at the protected resource. `validateToken()` returns the token scopes, but this path discards them and attaches the full API-key principal. Since requested and refresh-token scopes are not currently constrained to the client allowed scopes, a token containing only `mcp:read` can still invoke write operations through the MCP proxy. Please enforce the required scope before attaching authorization, and validate requested scopes during authorization and refresh.

packages/backend/src/services/mcp-oauth/plexus-idp-provider.ts
```diff
@@ -0,0 +1,573 @@
...
+      return this.handleAuthorizationCodeGrant(parsed.data, reply);
+    }
+
+    return this.handleRefreshTokenGrant(parsed.data, reply);
+  }
+
+  async validateToken(token: string): Promise<{ keyName: string; scopes: string[] } | null> {
+    if (!token.startsWith(ACCESS_TOKEN_PREFIX)) return null;
+
+    const record = await this.repo.getAccessToken(token);
```
mcowger, line 367: [P1] Validate the token resource at the MCP resource server. The token record stores `resource`, but `validateToken()` never checks it and the MCP hook calls validation with only the bearer token. This allows a token issued for another resource, or one issued before `mcpOAuth.resource` changes, to be accepted as valid MCP authentication. RFC 8707 resource validation must also happen at the protected resource. Please compare the stored resource with the current MCP resource before attaching API-key auth.

packages/backend/src/routes/management/mcp-oauth.ts
```diff
@@ -0,0 +1,37 @@
...
+  fastify.get('/v0/management/mcp-oauth/clients', async (_request, reply) => {
+    try {
+      const clients = await repo.listClientsWithActiveTokens();
+      return reply.send({ clients });
+    } catch (e: any) {
+      return reply.code(500).send({ error: e.message || 'Internal server error' });
+    }
+  });
+
+  fastify.post('/v0/management/mcp-oauth/tokens/:id/revoke', async (request, reply) => {
```
mcowger, line 93: [P1] Add client-level revocation. The management API only revokes individual token rows. There is no way to disable or delete an OAuth client, or revoke all of its access and refresh tokens in one operation. A compromised client therefore remains usable, and administrators cannot decommission it cleanly. Please add client lifecycle state and a client-level revoke or disable operation, and reject authorization and token requests for disabled clients.

packages/backend/src/services/mcp-oauth/plexus-idp-provider.ts
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    const code = randomToken(AUTH_CODE_PREFIX);
+    await this.repo.createAuthorizationCode({
+      code,
+      clientId: data.client_id,
+      redirectUri: data.redirect_uri,
+      resource: data.resource,
+      scope: data.scope ?? DEFAULT_SCOPES.join(' '),
+      keyName: authResult.keyName,
```
mcowger, line 329: [P1] Invalidate authorization codes when the API key rotates. Authorization codes store only `keyName`; during exchange, `issueToken()` resolves the current secret hash and binds the new token to whatever secret is currently configured. Key rotation revokes existing tokens but does not invalidate outstanding authorization codes, so a code granted under the old secret can still be exchanged after rotation to mint a fresh token under the replacement secret. Store the bound secret hash with the code, or revoke and validate pending codes as part of key rotation.

packages/backend/src/routes/mcp/index.ts
```diff
@@ -67,61 +157,48 @@ export async function registerMcpRoutes(
...
-  fastify.get('/.well-known/oauth-protected-resource', async (_request, reply) => {
-    logger.silly('OAuth protected resource discovery');
-    return reply.send({
-      resource: '/',
-      authorization_servers: ['/'],
-      scopes_supported: ['read', 'write'],
-      bearer_methods_supported: ['header', 'body', 'query'],
-    });
-  });
+  if (oauthEnabled && authProvider) {
```
mcowger: [P2] Keep discovery routes and the documented contract consistent. This branch only registers the authorization-server and protected-resource metadata routes when OAuth is enabled; the legacy `/.well-known/openid-configuration` endpoint is removed entirely. As a result, that endpoint returns 404 both with OAuth enabled and disabled, while `openapi.json` still advertises it, along with the other discovery and registration endpoints, as public compatibility routes. Please either preserve and update the discovery endpoints consistently or update the OpenAPI contract and compatibility expectations.

packages/backend/src/services/mcp-oauth/plexus-idp-provider.ts
```diff
@@ -0,0 +1,573 @@
...
+      clientId: input.clientId,
+      keyName: input.keyName,
+      apiKeySecretHash,
+      resource: input.resource,
+      scope: input.scope,
+      accessTokenExpiresAt: Date.now() + ACCESS_TOKEN_TTL_MS,
+      refreshTokenExpiresAt: Date.now() + REFRESH_TOKEN_TTL_MS,
+    });
+
+    return reply.send({
```
mcowger, line 514: [P2] Prevent token responses from being cached. OAuth token responses should include `Cache-Control: no-store` and `Pragma: no-cache`. This response contains both access and refresh tokens but does not set either header, leaving browser or intermediary caching behavior unspecified and potentially retaining replayable credentials. Please set these headers before sending the token response.

mcowger (commented):
Inline review comments for the requested MCP OAuth findings.

gwyntel-git:
Thanks for the thorough review — every single finding was valid. All 8 are addressed in the latest commit (3bb390d6).

**C1** — Routes always registered; each handler resolves `getMcpAuthProvider()` at request time. Auth hook unified into a single `onRequest` — no restart needed, reflects live config.

**C2** — `validateToken()` now checks the underlying API key's lifecycle via `isKeyDisabled()` after the secret-hash binding check. Expired/disabled/missing keys reject the token.

**C3** — Scopes enforced at the protected resource. The auth hook stores granted scopes on the request; a `preHandler` hook checks required scope per operation (`mcp:read` for GET/DELETE, `mcp:write` for `tools/call` POST). Insufficient scope → 403 with `WWW-Authenticate` header. Authorization and refresh also constrain requested scopes to the client's registered scope set.

**C4** — `validateToken()` compares the token's stored `resource` against the currently configured `mcpOAuth.resource` (RFC 8707). Rejects tokens minted for another resource or before the configured resource changed.

**C5** — Client lifecycle: `status` column (`active`/`disabled`) on `mcp_oauth_clients`, repository methods (`setClientStatus`/`deleteClient`/`revokeTokensForClient`), management API (`PATCH`/`DELETE`/`POST /clients/:clientId`). Authorization, token, and refresh endpoints all reject disabled clients.

**C6** — `apiKeySecretHash` stored with authorization codes at issuance. Code exchange verifies the stored hash matches the current API key secret. Key rotation invalidates outstanding codes — they cannot be exchanged post-rotation.

**C7** — Legacy `/.well-known/openid-configuration` restored (maps to authorization server metadata). When OAuth is disabled, all discovery endpoints return a consistent `404 oauth_disabled` instead of the old 404-or-canned-stub split. Route set now matches what `openapi.json` advertises.

**C8** — `Cache-Control: no-store` + `Pragma: no-cache` set on the token response before sending.

Typecheck passes, 885 unit tests pass, migration schema generates cleanly. Happy to adjust anything that doesn't land right.

packages/backend/src/routes/mcp/index.ts
gwyntel-git: Fixed in 3bb390d6. All OAuth routes (discovery, DCR, authorize, token) are now registered unconditionally. Each handler resolves the auth provider via `getMcpAuthProvider()` at request time — when MCP OAuth is disabled the provider is `null` and the endpoint returns 404 with an `oauth_disabled` error body. The MCP protected-route auth hook is unified into a single `onRequest` that handles both API-key and OAuth bearer tokens on every request, so toggling `mcpOAuth.enabled` in the Config UI takes effect immediately without a restart.

packages/backend/src/services/mcp-oauth/plexus-idp-provider.ts
gwyntel-git: Fixed in 3bb390d6. `validateToken()` now calls `isApiKeyCurrentlyValid(keyName)` after the secret-hash binding check. This uses the same `isKeyDisabled()` lifecycle check as `validatePlexusApiKey`, which rejects keys where `disabledAt` is set or `expiresAt` has passed. An OAuth access token is now invalidated the moment its underlying API key expires or is disabled.

packages/backend/src/routes/mcp/index.ts
gwyntel-git: Fixed in 3bb390d6. The OAuth fallback auth hook now stores granted scopes on the request (`mcpOAuthScopes`), and a `preHandler` hook enforces them: `tools/call` requires `mcp:write`, everything else requires `mcp:read`. Insufficient scope returns 403 with a `WWW-Authenticate: Bearer error="insufficient_scope"` header. During authorization and refresh, requested scopes are constrained to the client's registered scopes via `constrainScopes()` — clients cannot be granted scopes they weren't registered for.

packages/backend/src/services/mcp-oauth/plexus-idp-provider.ts
gwyntel-git: Fixed in 3bb390d6. `validateToken()` now calls `resourceMatchesCurrentConfig(resource)` which compares the token's stored resource against the current `mcpOAuth.resource` config value using RFC 8707 resource matching. Tokens minted for a different resource, or minted before the resource was reconfigured, are rejected.

packages/backend/src/routes/management/mcp-oauth.ts
gwyntel-git: Fixed in 3bb390d6 + 0a710491. Added a `status` column (`active`/`disabled`) to `mcp_oauth_clients`. New repository methods: `setClientStatus()`, `deleteClient()`, `revokeTokensForClient()`. New management API endpoints:
- `PATCH /v0/management/mcp-oauth/clients/:clientId` — toggle client status
- `DELETE /v0/management/mcp-oauth/clients/:clientId` — delete client
- `POST /v0/management/mcp-oauth/clients/:clientId/revoke` — revoke all tokens for a client

Authorization, token exchange, and refresh all reject disabled clients with `invalid_client` / `invalid_grant`. Admin UI added in 0a710491: status badge, disable/enable toggle, revoke-all-tokens, and delete buttons on each client card.

packages/backend/src/services/mcp-oauth/plexus-idp-provider.ts
gwyntel-git: Fixed in 3bb390d6. Authorization codes now store the `apiKeySecretHash` that was current at issuance time. During code exchange, the stored hash is validated against the current API key secret via `isTokenBoundToCurrentApiKeySecret()`. If the API key has been rotated since the code was issued, the exchange fails with `invalid_grant` and the code is consumed.

packages/backend/src/routes/mcp/index.ts
gwyntel-git: Fixed in 3bb390d6. The legacy `/.well-known/openid-configuration` endpoint is restored and returns the same RFC 8414 authorization server metadata as `/.well-known/oauth-authorization-server`. All discovery endpoints (`oauth-authorization-server`, `oauth-protected-resource`, `openid-configuration`) are always registered and return 404 with an `oauth_disabled` error when MCP OAuth is off, matching the `openapi.json` contract.

packages/backend/src/services/mcp-oauth/plexus-idp-provider.ts
gwyntel-git: Fixed in 3bb390d6. The token endpoint now sets `Cache-Control: no-store` and `Pragma: no-cache` before sending the response, preventing browsers and intermediaries from caching access/refresh tokens.

mcowger:
Closing in favor of an updated PR incorporating review fixes, client lifecycle UI, and scoped browser authorization flow.

mcowger:
I would have pushed via this PR, but this one wasn't marked with maintiner edits allowed.

Issue #723, opened 2026-07-14
Claude Code masking: stale CC tool reference renames genuine current-CC tools to mcp__*


mcowger:
## Summary

Genuine, current Claude Code tools are being collision-renamed to `mcp__*` by the OAuth masking pipeline because the CC tool reference (`cc-reference-tools.ts`) is stale relative to the Claude Code version real clients send. A real Claude Code request would never carry `mcp__EnterWorktree` etc., so this is a **fingerprint-fidelity** issue for the masked outbound request.

Discovered while validating the native Anthropic OAuth path (NOMOV3 M1) against a real `claude-cli/2.1.207` capture. Reversal handles the renamed names correctly, so this is **not** a correctness bug in request/response tool naming — it only affects how faithfully our masked request mimics real Claude Code.

## Evidence

Real capture: `User-Agent: claude-cli/2.1.207 (external, cli)`.

Feeding its genuine tool set through `applyClaudeCodeMasking`, `ccCollisionShape.detect()` renames three legitimate CC tools:

```
["EnterWorktree","mcp__EnterWorktree", "ALWAYS USE THIS TOOL INSTEAD OF EnterWorktree."]
["ScheduleWakeup","mcp__ScheduleWakeup","ALWAYS USE THIS TOOL INSTEAD OF ScheduleWakeup."]
["Workflow","mcp__Workflow", "ALWAYS USE THIS TOOL INSTEAD OF Workflow."]
```

## Root cause (two contributing factors)

**1. `matchesReferenceShape` treats "no required params" as a mismatch.**
In CC 2.1.207 these tools declare **no `required` key** (all params optional):

| tool | golden `required` | reference (`cc-reference-tools.ts`) |
|---|---|---|
| `EnterWorktree` | absent | `[]` |
| `Workflow` | absent | `[]` |
| `ScheduleWakeup` | absent | `['delaySeconds','reason','prompt']` |

`requiredParamsOf()` returns `undefined` for an absent `required`, and `matchesReferenceShape()` has:

```ts
if (!requiredParams) return false;
```

So `EnterWorktree`/`Workflow` mismatch **even though their reference is `[]`** — "no required params" should compare equal to a reference of `[]`, but the guard returns false first.

**2. The reference shapes/version are stale.**
[...]

PR #795, opened 2026-08-06
fix(oauth): stop double-prefixing mcp__ tools and cover tool_reference blocks in masking renames

kaspesi:
## Summary

With Claude-Code masking active on the Anthropic OAuth path, any Claude Code session that has MCP servers connected fails with a `400` from Anthropic on **every** turn:

```json
{"type":"error","error":{"type":"invalid_request_error","message":"Tool reference 'mcp__linear-server__get_issue' not found in available tools"}}
```

The failure is sticky. The block Anthropic objects to lives in the client's persisted message history, so once the session hits it, every subsequent turn fails identically until the user clears the conversation. We observed 82 such failures in a single day across several different MCP servers before applying this fix.

---

## Root cause

Two independent defects that only produce a visible failure when combined.

### Defect 1 — `mcp-shape` double-prefixes tools already in `mcp__<server>__<tool>` form

`transformers/oauth/masking/mcp-shape.ts` normalizes opencode-style flat `<server>_<tool>` names into Claude Code's `mcp__<server>__<tool>` convention. It detects a "server" by clustering tool names on the substring before their **first** underscore, and renames any cluster with `MIN_CLUSTER_SIZE` (4) or more members.

Names that are *already* in `mcp__<server>__<tool>` form — which is exactly what real Claude Code sends — have `mcp` as their first-underscore prefix. Any real Claude Code session with an MCP server connected has well over four such tools, so the entire set clusters under a phantom server literally named `mcp`, and every tool gets prefixed a second time.

### Defect 2 — renames never reached `tool_reference` blocks

`transformers/oauth/masking/rename-apply.ts` applied renames to `tools[].name`, `tool_choice.name` and `tool_use.name`, but not to `tool_reference` blocks. Those blocks are produced by Anthropic's advanced tool use / tool search feature (the `tool_search_tool_bm25_20251119` / `tool_search_tool_regex_20251119` server tools — this repo already knows about them in `transformers/adapters/strip-unsupported-tool-search.adapter.ts`) and are persisted in the client's message history. `reverse-rename.ts` had the mirror-image gap: it reversed the `"name"` JSON key but not `"tool_name"`.

So `tools[]` was rewritten while the `tool_reference.tool_name` blocks referring to those same tools were not. Anthropic validates the two against each other and rejects the mismatch.

Note this is **v1/v2 parity**, not a new capability: `transformers/oauth/oauth-claude.ts` (the v1 path) already handles `tool_reference` blocks, including ones nested inside `tool_result.content[]` (see the `case 'tool_reference'` and `case 'tool_result'` arms of `applyClaudeOAuthTransform`). The v2 masking pipeline simply never grew the same coverage.

Defect 1 alone would be a fidelity problem; defect 2 alone is latent until something genuinely gets renamed. Together they are a hard failure on the default configuration.

---

## Before / after

Feeding a realistic Claude Code tool list through `buildToolRenamePairs()`:

| Tool name sent by Claude Code | Rename emitted **before** | Rename emitted **after** |
|---|---|---|
| `mcp__linear-server__get_issue` | `mcp__mcp___linear-server__get_issue` | *(none — left alone)* |
| `mcp__playwright__browser_click` | `mcp__mcp___playwright__browser_click` | *(none — left alone)* |
| `mcp__github__get_me` | `mcp__mcp___github__get_me` | *(none — left alone)* |
| `github_search_users` (flat, opencode-style) | `mcp__github__search_users` | `mcp__github__search_users` *(unchanged behaviour)* |

Note the triple underscore in the mangled form: `mcp__` + `mcp` + `___linear-server...`.

The resulting request body is internally inconsistent:

```jsonc
{
  "tools": [ { "name": "mcp__mcp___linear-server__get_issue" } ],   // renamed
  "messages": [
    { "role": "user", "content": [
      { "type": "tool_result", "tool_use_id": "toolu_...", "content": [
        { "type": "tool_reference", "tool_name": "mcp__linear-server__get_issue" }  // NOT renamed
      ]}
    ]}
  ]
}
```

→ `400 Tool reference 'mcp__linear-server__get_issue' not found in available tools`.

---

## Reproduction

Against `main` (no patch), in `packages/backend`:

```ts
import { buildToolRenamePairs } from './src/transformers/oauth/masking/registry';

const names = [
  'mcp__linear-server__get_issue',
  'mcp__linear-server__save_issue',
  'mcp__playwright__browser_click',
  'mcp__playwright__browser_navigate',
];
console.log(buildToolRenamePairs(names.map((name) => ({ name }))));
```

Before this PR, all four are renamed to `mcp__mcp___*`. After, the result is `[]`.

End to end: point a current Claude Code client at a Plexus provider on the Anthropic OAuth path with masking active, connect any MCP server exposing four or more tools, and let tool search fire once. The first turn after the tool-search result lands fails with the 400 above, and so does every turn after it.

---

## What the fix changes

| File | Change |
|---|---|
| `masking/mcp-shape.ts` | Added an `ALREADY_MCP_SHAPED` guard (`/^mcp__[^_].*__.+$/`). Tools already in the target convention are skipped before clustering, so they are never counted toward a cluster nor renamed. Flat `<server>_<tool>` clustering is untouched. |
| `masking/rename-apply.ts` | `applyToolRenames()` now also renames `tool_reference.tool_name`, both at the top level of a message's `content[]` and nested inside a `tool_result`'s `content[]` (the shape tool-search results arrive in). Existing `tool_use` / `tools[]` / `tool_choice` handling is unchanged. |
| `masking/reverse-rename.ts` | `reverseToolRenames()` now reverses the `tool_name` key in addition to `name`, in both the plain and backslash-escaped-quote forms. The key-scoped matching (rather than bare substring) is preserved. |

The change is additive and behaviour-preserving for every input that worked before: flat-prefixed opencode tool sets still cluster and rename exactly as they did.

---

## Test coverage added

All new tests were verified non-vacuous — reverting the three source files to their `main` versions makes **5 of the 26** tests in these files fail.

- `masking/__tests__/registry.test.ts` (+2)
  - CC-shaped `mcp__<server>__<tool>` tools produce no rename pairs at all.
  - Flat `<server>_<tool>` names still cluster and rename correctly when CC-shaped MCP tools are present alongside them (guards against over-correcting).
- `masking/__tests__/rename-apply.test.ts` (new file, 3 tests)
  - A top-level `tool_reference` block is renamed in lockstep with `tools[]`.
  - `tool_reference` blocks nested inside a `tool_result.content[]` are renamed, and unrelated tool names inside the same array are left alone.
  - `tool_use` renaming still works and free-text `text` blocks that happen to contain the tool name are not touched.
- `masking/__tests__/reverse-rename.test.ts` (+2)
  - `tool_reference.tool_name` is reversed on the response path.
  - A renamed name appearing as some *other* `*_name` field's value (e.g. `"server_name":"Bash"`) is still **not** rewritten — the key scoping stays exact.

Verification run on this branch:

- Full backend suite green: **2722 tests / 242 files** (`bun run test:force-all` in `packages/backend`).
- `bun run typecheck`, `biome format`, `biome lint` all clean (the full lefthook pre-commit chain passes).

No schema or migration changes.

---

## Why this matters

Tool search is on by default in recent Claude Code, and MCP servers are the common case rather than an edge case. Any user pointing Claude Code at a Plexus provider on the masked OAuth path therefore hits this on the first turn after tool search fires — and because the offending `tool_reference` block is persisted client-side, the session never recovers on its own. From the user's side it looks like the gateway has simply stopped working, with an error message naming a tool they can see is present in their config.

---

## Related

- [#775](https://github.com/mcowger/plexus/pull/775) (merged) fixed two other defects on this same OAuth path — dropped `anthropic-beta` flags and `description` injection into server-side tools. This is the third defect found in that same investigation; it was masked by those two until they were fixed, since requests were failing earlier in the pipeline. Same area of the code, independent root cause.
- [#723](https://github.com/mcowger/plexus/issues/723) (closed) covers a *different* over-renaming defect in the same masking pipeline — `cc-collision-shape` renaming genuine current-CC tools to `mcp__*` due to a stale reference set. Adjacent symptom, unrelated cause; this PR does not touch `cc-collision-shape`.

I did not find an existing open issue for this specific failure.

### DIFF
--- a/packages/backend/src/transformers/oauth/masking/__tests__/registry.test.ts
+++ b/packages/backend/src/transformers/oauth/masking/__tests__/registry.test.ts
@@ -62,6 +62,37 @@ describe('buildToolRenamePairs', () => {
     });
   });
 
+  it('leaves tools that are already in mcp__<server>__<tool> form untouched', () => {
+    // Real Claude Code already emits this convention. Clustering on the first
+    // underscore reads their literal "mcp" prefix as a server name and
+    // double-prefixes them (mcp__github__get_me ->
+    // mcp__mcp___github__get_me), which desynchronizes tools[].name from the
+    // tool_reference blocks in the message history and makes Anthropic reject
+    // the request with 400 "Tool reference ... not found in available tools".
+    const tools = [
+      tool('mcp__github__get_me'),
+      tool('mcp__github__search_users'),
+      tool('mcp__playwright__browser_click'),
+      tool('mcp__playwright__browser_navigate'),
+      tool('mcp__home-assistant__ha_get_state'),
+    ];
+    expect(buildToolRenamePairs(tools)).toEqual([]);
+  });
+
+  it('still clusters flat server_tool names when CC-shaped MCP tools are present alongside them', () => {
+    const tools = [
+      ...Array.from({ length: 4 }, (_, i) => tool(`mcp__playwright__tool_${i}`)),
+      ...Array.from({ length: 4 }, (_, i) => tool(`github_tool_${i}`)),
+    ];
+    const pairs = buildToolRenamePairs(tools);
+    expect(Object.fromEntries(pairs.map(([from, to]) => [from, to]))).toEqual({
+      github_tool_0: 'mcp__github__tool_0',
+      github_tool_1: 'mcp__github__tool_1',
+      github_tool_2: 'mcp__github__tool_2',
+      github_tool_3: 'mcp__github__tool_3',
+    });
+  });
+
   it('does not cluster prefixes with too few tools (avoids false positives)', () => {
     // Only 3 tools share the "list" prefix — below MIN_CLUSTER_SIZE (4) — so
     // opencode's own list_mcp_resources / list_mcp_resource_templates /
--- a/packages/backend/src/transformers/oauth/masking/__tests__/rename-apply.test.ts
+++ b/packages/backend/src/transformers/oauth/masking/__tests__/rename-apply.test.ts
@@ -1,0 +1,83 @@
+/**
+ * Regression tests for `applyToolRenames`'s handling of `tool_reference`
+ * blocks (Anthropic advanced tool use / tool search).
+ *
+ * A renamed tool must be renamed everywhere its name appears in the request,
+ * including the `tool_reference.tool_name` blocks a tool-search result leaves
+ * in the message history. If `tools[]` is renamed but those blocks are not,
+ * Anthropic rejects the whole request with
+ * `400 Tool reference '<name>' not found in available tools` — and because
+ * the block is persisted in the client's history, every subsequent turn in
+ * that session fails the same way.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { applyToolRenames } from '../rename-apply';
+import type { RenamePair } from '../types';
+
+const pairs: RenamePair[] = [['github_search_users', 'mcp__github__search_users']];
+
+describe('applyToolRenames', () => {
+  it('renames a top-level tool_reference block', () => {
+    const body = {
+      tools: [{ name: 'github_search_users' }],
+      messages: [
+        {
+          role: 'assistant',
+          content: [{ type: 'tool_reference', tool_name: 'github_search_users' }],
+        },
+      ],
+    };
+    const result = applyToolRenames(body, pairs);
+    expect(result.messages[0].content[0]).toEqual({
+      type: 'tool_reference',
+      tool_name: 'mcp__github__search_users',
+    });
+    expect(result.tools[0].name).toBe('mcp__github__search_users');
+  });
+
+  it('renames tool_reference blocks nested inside a tool_result (tool-search result shape)', () => {
+    const body = {
+      messages: [
+        {
+          role: 'user',
+          content: [
+            {
+              type: 'tool_result',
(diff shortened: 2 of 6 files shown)


mcowger:
Thanks @kaspesi — this is an excellent writeup and the fix is solid. I validated the description claims directly against the branch before merging:

- Reproduced the double-prefixing on `main` exactly as described (`mcp__linear-server__get_issue` → `mcp__mcp___linear-server__get_issue`), and confirmed the guard eliminates it on this branch.
- Confirmed the `tool_reference` gap on `main` in both `rename-apply.ts` and `reverse-rename.ts`, and the v1 parity claim about `oauth-claude.ts`.
- Verified the test non-vacuity claim: reverting the three source files makes exactly 5 of 26 tests fail in the named files.
- Behaviour preservation holds — flat `<server>_<tool>` clustering still works alongside CC-shaped MCP tools.
- Re-ran verification myself: full backend suite green (2722/242) and typecheck clean.

Two non-blocking nits for a follow-up if you're inclined: the module headers in `reverse-rename.ts` ("scoped to the `\"name\"` key") and `rename-apply.ts` ("only renames the `name` field") are now slightly stale since both also cover `tool_name`.

Third OSS contribution landed from that same real-world OAuth investigation ("we observed 82 such failures in a single day") — much appreciated.

PR #806, opened 2026-08-13
feat: alias-as-fallback-target

jarjarthidev:
## Summary

Adds support for using one model alias as a fallback target for another alias.

### Backend

- Extends alias targets to support either:
  - A concrete `{ provider, model }` target, or
  - An `{ alias }` reference, but not both.
- Persists alias references in `model_alias_targets.target_alias_slug`, while allowing `provider_slug` and `model_name` to be nullable.
- Expands nested alias references during routing and preserves the referenced alias’s target groups and selectors.
- Treats alias references as fallback candidates after selector-ordered concrete targets.
- Skips disabled alias references and deduplicates candidates resolving to the same provider/model.
- Adds cycle detection during configuration loading, alias updates, and configuration rebuilds, with nickname-aware resolution.
- Excludes alias references from provider/model metadata discovery and background exploration.
- Adds startup repair for SQLite databases affected by the original table-recreation migration, restoring corrupted concrete targets without modifying legitimate alias references.

### Frontend

- Adds an Alias/Model target selector to the target group editor.
- Allows selecting another alias as a fallback target while excluding the alias currently being edited.
- Displays alias targets with a link indicator and independent enable/disable controls in desktop and mobile alias views.
- Updates API types, serialization, model lists, deduplication, and metadata handling for targets without provider/model fields.

### Migrations and tooling

- Adds SQLite and PostgreSQL migrations for the new `target_alias_slug` column and nullable concrete-target fields.
- Corrects the SQLite migration to populate the new column with `NULL` for existing concrete targets.
- Updates migration generation and linting to detect unsafe SQLite table-recreation `INSERT ... SELECT` statements that reference columns absent from the source table, rewriting them to `NULL` and failing validation if unsafe references remain.

### Tests

- Adds routing coverage for direct, nested, disabled, ordered, deduplicated, and cyclic alias references.
- Adds migration integrity and corruption-repair regression tests.
- Adds frontend tests for alias-target deduplication and provider/model list handling.

### DIFF
--- a/packages/backend/src/config.ts
+++ b/packages/backend/src/config.ts
@@ -713,11 +713,19 @@ export const ProviderConfigSchema = z
     message: 'raw_passthrough currently supports static API-key providers only',
   });
 
-const ModelTargetSchema = z.object({
-  provider: z.string(),
-  model: z.string(),
-  enabled: z.boolean().default(true).optional(),
-});
+const ModelTargetSchema = z
+  .object({
+    provider: z.string().optional(),
+    model: z.string().optional(),
+    alias: z.string().min(1).optional(),
+    enabled: z.boolean().default(true).optional(),
+  })
+  .refine(
+    (data) => (data.alias ? !data.provider && !data.model : !!data.provider && !!data.model),
+    {
+      message: "A target must specify either 'alias' or both 'provider' and 'model', but not both",
+    }
+  );
 
 const SelectorTypeSchema = z.enum([
   'random',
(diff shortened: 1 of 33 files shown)


mcowger (commented):
Found three functional routing/configuration issues; comments below.

packages/backend/src/services/routing/router.ts
```diff
@@ -310,11 +313,111 @@ async function selectOrderedTargets(
...
+  // of the ordered concrete results, then walk group.targets in their
+  // declared order so alias-ref expansions land in their author-declared
+  // position relative to concrete targets, instead of always trailing.
+  const resultsByKey = new Map<string, RouteResult>();
+  for (const result of results) {
+    resultsByKey.set(`${result.provider}\u0000${result.model}`, result);
+  }
+
+  const merged: RouteResult[] = [];
+  for (const target of group.targets) {
```
mcowger, line 379: [P0] Preserve selector order when merging alias refs. `selectOrderedTargets()` has already produced the selector-defined order in `results`, but iterating `group.targets` here and looking up by key reverts all concrete candidates to declaration order. That makes cost/random/latency/performance/usage selectors route the wrong first target in both `resolveCandidates()` and `resolve()`/direct-group routing. Merge alias expansions into the selector-ordered concrete stream instead.

packages/backend/src/services/routing/router.ts
```diff
@@ -310,11 +313,111 @@ async function selectOrderedTargets(
...
+        );
+        continue;
+      }
+      const nested = await Router.resolveCandidates(
+        target.alias,
+        incomingApiType,
+        sessionKey,
+        visited
+      );
+      merged.push(...nested);
```
mcowger: [P0] Preserve the outer alias metadata on nested candidates. `nested` carries the referenced alias's `canonicalModel`/`incomingModelAlias`, and pushing it unchanged means a request through `A -> B` is treated as B for alias-level `extraBody`/advanced behaviors, context limits, vision fallthrough, compaction, and sticky-session recording. Expanded candidates should retain A's request metadata while using B only for its target-group expansion and concrete route data.

packages/backend/src/config.ts
```diff
@@ -1252,6 +1322,8 @@ export function validateConfig(configJson: string): PlexusConfig {
 }
 
 function hydrateConfig(config: z.infer<typeof RawPlexusConfigSchema>): PlexusConfig {
+  assertNoAliasRefCycles(config.models);
```
mcowger, line 1327: [P1] Apply cycle validation to the DB-backed configuration path. This check is only reached by `validateConfig()`/`hydrateConfig()`, while the running service loads aliases directly in `ConfigService.doRebuild()` and management PUT/PATCH saves only run `ModelConfigSchema.safeParse()`. A cycle can therefore be persisted through the UI/API and later becomes an unroutable alias instead of being rejected at configuration time; validate the complete repository model graph before accepting/rebuilding it.

mcowger (commented):
The routing and configuration findings above are blocking: selector ordering and nested alias metadata are P0; DB cycle validation is P1.

packages/backend/src/services/routing/router.ts
mcowger: Already fixed as of `1dfdda9c`. `buildGroupCandidates` now computes `results` via `selectOrderedTargets(group.selector, enriched)` (router.ts:353) and seeds `merged` from that selector-ordered array (router.ts:377) — concrete candidates are never rebuilt from `group.targets` declaration order. The `group.targets` loop (router.ts:379-407) only appends alias-ref expansions; it doesn't touch concrete ordering. Verified with a cost-selector test where the cheaper `p2/m2` target routes before `p1/m1`. Resolving.

packages/backend/src/services/routing/router.ts
mcowger: Already fixed as of `1dfdda9c`/`db172e12`: nested candidates are rewritten with `canonicalModel`/`incomingModelAlias` set to the outer alias (router.ts:403-404), and `modelArchitecture` falls back to the nested candidate's own architecture when the outer alias has none (router.ts:401-402, `?? candidate.modelArchitecture`). Resolving.

packages/backend/src/config.ts
mcowger: Already fixed. `ConfigService.doRebuild()` calls `assertNoAliasRefCycles(models)` against the full repository-loaded graph before rebuilding the cache (config-service.ts:472), and the management PUT/PATCH alias handlers call `assertNoAliasRefCycles({ ...currentModels, [slug]: result.data })` against the merged full graph before `saveAlias()` persists (config.ts:299, config.ts:337). Resolving.

packages/backend/src/config.ts
```diff
@@ -752,6 +764,64 @@ export function findDuplicateAliasTargets(
...
+    }
+  }
+
+  const referencedAliases = (slug: string): string[] => {
+    const model = models[slug];
+    if (!model?.target_groups) return [];
+    const refs: string[] = [];
+    for (const group of model.target_groups) {
+      for (const target of group.targets) {
+        if (target.enabled !== false && target.alias) {
```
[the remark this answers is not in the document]
mcowger: This is intentional: `assertNoAliasRefCycles` skipping `enabled: false` edges mirrors the router's own expansion logic, which already skips disabled alias-ref targets (router.ts:381, `if (target.enabled === false) continue;`). A disabled `A -> B` edge can never be traversed at request time, so it can't produce a live routing cycle, and the config-validation false positive was blocking legitimate configs (e.g. two aliases with a mutual fallback where one leg is temporarily disabled).

On "hiding" a cycle: re-enabling a disabled target always goes through a path that re-runs `assertNoAliasRefCycles` against the full graph — `ConfigService.doRebuild()` on every config load, and the management PUT/PATCH alias handlers on every save (see the reply on the P1 thread above). There's no way to flip a target back to `enabled: true` without re-validating the complete alias graph, so a cycle can't become live without being caught at that point.

Keeping this as-is.

PR #865, opened 2026-09-09
feat(models): optional auto-alias creation for autosynced provider models

kaspesi:
## Problem

`model_autosync` keeps a provider's `models` map current, but those models are **not routable**: inference resolves the request `model` against **aliases** only (`Router.resolveCandidates`/`resolve` → `findAlias`, which reads `config.models`), and `GET /v1/models` enumerates aliases only. So a newly-discovered provider model never appears in `/v1/models` and can't be called by bare name until an operator hand-creates an alias for it. Autosync fills half the gap (the provider `models` map) but not the half clients actually use.

## Change

Add an opt-in `model_autosync.createAliases` flag (default `false`). When enabled, after autosync inserts newly-discovered models into the provider's `models` map, the scheduler also creates a **passthrough alias** (slug === model id) for each new model that has **no existing alias**:

```jsonc
{
  "priority": "api_match",
  "sticky_session": true,
  "target_groups": [
    { "name": "<provider>", "selector": "random",
      "targets": [{ "provider": "<provider>", "model": "<id>", "enabled": true }] }
  ]
}
```

- **Never overwrites** an alias that already exists (a slug already taken by this or any other provider is left untouched).
- Fires the existing models-changed callback when aliases *or* provider models were added, so the change is picked up without a restart.
- `preferred_api` is intentionally omitted so the model family determines the default surface.

## Details

- New `model_autosync_create_aliases` column (sqlite + postgres schema). Migrations not committed per repo workflow — CI regenerates on merge.
- `ModelAutosyncScheduler.createMissingAliases()` + wiring in `runSyncNow()`.
- OpenAPI `ProviderConfig.model_autosync.createAliases` documented; asset regenerated.
- Admin UI: "Auto-create model aliases" checkbox in the provider Advanced editor (disabled unless autosync is enabled).
- Tests: scheduler create / skip-existing / off / callback-on-alias-only + DB round-trip of the flag.

### Drive-by fix

Set an explicit `Connection: keep-alive` header on MCP SSE responses instead of relying on a runtime-implicit header. This fixes a pre-existing `mcp-routes` SSE test (`expect(response.headers.connection).toBe('keep-alive')`) that fails under Bun's `inject`/light-my-request harness, where the implicit header isn't synthesized. Needed to get a green `bun run test` locally.

## Test plan

- `bun run typecheck` — all workspaces pass
- `bun run test` — full backend suite green (incl. new scheduler + DB tests)
- Manually verified on a live gateway: enabling the flag surfaces a newly-catalogued Codex model in `/v1/models` and routes an inference call to it end-to-end.

### DIFF
--- a/docs/openapi/components/schemas/ProviderConfig.yaml
+++ b/docs/openapi/components/schemas/ProviderConfig.yaml
@@ -267,6 +267,15 @@ properties:
         type: integer
         minimum: 1
         default: 60
+      createAliases:
+        type: boolean
+        default: false
+        description: >
+          When true, autosync also creates a passthrough model alias (slug
+          equal to the model id) for each newly discovered provider model that
+          has no existing alias, so the model becomes routable and appears in
+          GET /v1/models without a manually created alias. Never overwrites an
+          alias that already exists.
   adapter:
     oneOf:
       - type: string
--- a/packages/backend/drizzle/schema/postgres/providers.ts
+++ b/packages/backend/drizzle/schema/postgres/providers.ts
@@ -41,6 +41,7 @@ export const providers = pgTable(
     quotaCheckerOptions: text('quota_checker_options'), // JSON or encrypted string
     modelAutosyncEnabled: boolean('model_autosync_enabled').notNull().default(false),
     modelAutosyncInterval: integer('model_autosync_interval').notNull().default(60),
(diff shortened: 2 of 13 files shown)


mcowger:
Why not simply using the direct/ prefix?

PR #866, opened 2026-09-09
feat(oauth-masking): opt-in preserveToolDescriptions to keep tool descriptions

kaspesi:
## Problem

The Claude Code OAuth-masking pipeline (`transformers/oauth/masking/cc-tools.ts` → `stripDescriptionsAndInjectSyntheticTools`) blanks **every** custom/MCP tool `description` to `''` before the request reaches Anthropic, for fingerprint parity — the doc's stated reason is *"real Claude Code sessions send minimal/no tool descriptions."*

I captured a genuine Claude Code session on the wire (v2.1.246) and that premise doesn't hold:

| tool class | real CC `description` |
| --- | --- |
| MCP tools (user-configured) | full text (e.g. 45 / 105 chars) |
| built-in tools | large — `Bash` ≈ 9,822 chars, `Agent` ≈ 6,926, `Workflow` ≈ 3,480 |
| empty descriptions | **0 of 22 tools** |

Two consequences of blanking:
1. **Capability** — the model loses the info it needs to select/parameterize a tool. Repro: an MCP tool with an opaque name whose meaning lived only in its description — tool-search surfaced it, but the model **refused to call it** (*"no description of what it actually does"*). Any tool whose param semantics live in prose is similarly degraded.
2. **Fingerprint** — since genuine CC never sends an empty tool description, blanking arguably makes traffic *less* CC-like, not more.

## Change

Adds an **opt-in** config flag `oauthMasking.preserveToolDescriptions` (default `false`), threaded:

```
oauth-native-request.ts → applyClaudeCodeMasking(payloadStr, { preserveToolDescriptions })
                        → stripDescriptionsAndInjectSyntheticTools(body, renamePairs, options)
```

- **Default (`false`)** — unchanged legacy behavior (blank descriptions).
- **`true`** — keep the caller's real descriptions verbatim; collision renames get their disambiguation note **appended** to the real text instead of replacing it. Server-side (`type !== "custom"`) tools stay byte-identical in both modes.

Fully **non-breaking**: existing configs/callers are unaffected unless they opt in.

### Files
- `config.ts` — `OAuthMaskingConfigSchema` + optional `oauthMasking` field.
- `transformers/oauth/masking/cc-tools.ts` — `ToolMaskingOptions`, preserve logic, corrected module doc.
- `transformers/oauth/masking/apply-masking.ts` — thread the option through.
- `services/oauth/oauth-native-request.ts` — read the flag defensively and pass it in.
- Tests — new `cc-tools-preserve.test.ts` (both modes, server-tool passthrough, collision-note append/fallback) + end-to-end wiring assertions in `apply-masking.test.ts`.

### Verification
- `bun run typecheck`: pass. OAuth masking unit tests: 67/67 pass.
- Heads-up (not caused by this PR): `bun run test` currently has one failing test on `main`, `routes/mcp/__tests__/mcp-routes.test.ts` ("sets portable SSE cache headers", `connection: keep-alive` — looks environment-sensitive). This diff doesn't touch `routes/mcp`.

## Question for maintainer

The real judgment call is the **fingerprint tradeoff**, which you own. The evidence above suggests preserving descriptions is both more authentic and more capable, but I've kept it behind an opt-in flag (default off) so nothing changes for anyone unless they enable it. Happy to adjust in any direction you prefer, e.g.:
- scope preservation to `mcp__*`/server-provided tools only (leave CC-builtin-name tools as today), or
- flip the default once you're comfortable.

Landing this as an opt-in first lets instances that run genuine Claude Code (where the caller's descriptions already *are* CC's) get correct tool usage without changing default behavior.

### DIFF
--- a/.repomap.txt
+++ b/.repomap.txt
@@ -1999,7 +1999,7 @@
   │ [function] sortedUnique (line 64)
 
 📁 packages/backend/src/transformers/oauth/masking/cc-tools.ts:
-  │ [function] stripDescriptionsAndInjectSyntheticTools (line 49)
+  │ [function] stripDescriptionsAndInjectSyntheticTools (line 50)
 
 📁 packages/backend/src/transformers/oauth/masking/dedupe.ts:
   │ [function] dedupeSyntheticToolCollisions (line 28)
--- a/packages/backend/src/config.ts
+++ b/packages/backend/src/config.ts
@@ -792,6 +792,18 @@ export const McpOAuthConfigSchema = z.object({
   issuer: z.string().url().optional(),
 });
 
+const OAuthMaskingConfigSchema = z.object({
+  /**
+   * Preserve the caller's real tool `description` fields on the outbound
+   * Claude-Code-masked request instead of blanking them. Default `false`
+   * keeps the historical behavior. See
+   * `transformers/oauth/masking/cc-tools.ts` for the rationale (genuine
+   * Claude Code sends full tool descriptions, so preserving is both more
+   * authentic and lets the model use tools correctly).
+   */
+  preserveToolDescriptions: z.boolean().default(false),
+});
+
 const RawPlexusConfigSchema = z
   .object({
     providers: z.record(z.string(), ProviderConfigSchema),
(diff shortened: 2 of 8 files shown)


kaspesi:
@mcowger is there a reason that tool descriptions are redacted that I'm missing?

mcowger:
> @mcowger is there a reason that tool descriptions are redacted that I'm missing?

As noted previous checks didn't have them, but sounds like the behavior has changed.  

I think we should take this one as default though?

PR #871, opened 2026-09-10
feat(runtime): drain work during graceful shutdown

darkspadez:
Plexus could close storage while admitted request handlers, scheduled work, local process cleanup, or telemetry writes were still running. Active SSE responses could also keep shutdown open indefinitely.

This change adds one idempotent shutdown path for OS signals and management restarts. It closes active responses, drains admitted handlers and background producers, stops local MCP processes, flushes telemetry, and closes storage last under a bounded deadline. Schedulers and storage services gain terminal shutdown behavior so late asynchronous work cannot restart timers or write after storage closes.

The branch also passes the existing request ID into the two image-route disconnect trackers, restoring compatibility with the current upstream timeout API after rebasing.

Validation:
- Full backend suite: 3,216 of 3,217 tests passed
- The remaining Bun-specific MCP Connection header assertion is addressed separately in #870
- Image route suite: 9 tests passed
- `bun run typecheck`
- `bun run format`
- `bunx biome lint .`

### DIFF
--- a/packages/backend/src/index.ts
+++ b/packages/backend/src/index.ts
@@ -44,6 +44,7 @@ import { BackgroundExplorer } from './services/routing/background-explorer';
 import { CooldownManager } from './services/runtime/cooldown-manager';
 import { DebugManager } from './services/observability/debug-manager';
 import { ModelMetadataManager } from './services/models/model-metadata-manager';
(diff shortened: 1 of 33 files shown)


mcowger:
Closed as overlapping #870 870

PR #904, opened 2026-09-20
feat(oauth): replace hand-rolled muse-code with pi-ai meta provider

mcowger:
## Summary
The PR upgrades `@earendil-works/pi-ai` to version `0.86.1` across all workspaces and replaces the manually maintained Muse OAuth implementation with the newly available native `meta` provider from pi‑ai. The changes span both backend and frontend codebases:

- Updated TypeScript, lock files and workspace definitions to reflect the new version.
- Removed the legacy `muse-code.ts` service and its extensive test suite.
- Replaced all references to the old `muse-code` provider with `meta` in routes, adapters, resolver logic, and OAuth provider registries.
- Adjusted route URLs and handler functions to use the `meta` provider instead of `muse-code`.
- Updated front‑end components that referenced the old provider in UI rendering, form handling and credential management.
- Modified tests targeting the now‑deprecated `muse-code` provider to reference `meta` instead.
- Adjusted utility code that resolved OAuth base URLs and provider IDs to use the new identifier.
- Added a conventional commit with the commit hash `b8f5c8e6` for the migration.

## Changes
(See diff for exact file modifications; summary above covers all affected files.)

## Testing
Not run.


mcowger:
Re: the missing-migration concern — this is a deliberate decision, not an oversight.

The `muse-code` provider was never released beyond staging, so there is no production state to migrate. For staging, we're doing a manual cutover instead of a startup migration:

- Provider rows load unvalidated, so stale `oauth_provider: 'muse-code'` configs don't break boot — affected routes fail per-request with a clear error until re-pointed at `meta`.
- Stale checkers warn per poll and keep serving their last good snapshot (no paging, no crash).
- Re-login is required regardless: the old credential shape (`access` = JSON blob, empty `refresh`) is incompatible with pi's shape (`access` = minted key, `refresh` = identity token), so no migration could carry sessions over anyway.

This PR also wires the previously API-only credential delete into the provider OAuth editor (two-click confirm), so staging cleanup is click-through. Cutover steps: flip providers to `meta`, delete old `muse-code` credentials, log in to `meta` once per account.

PR #912, opened 2026-09-21
fix(frontend): clarify Auto Compat copy and gate toggles on pi-ai mapping

mcowger:
## Summary

Reworks the Auto Compat help text in the provider and model editors so it describes what the toggle actually does (translate reasoning/generation options into the upstream dialect using the mapped pi-ai model) instead of implying it *is* the pi-ai provider selection. Both toggles are now disabled unless a pi-ai Provider is selected, since without the mapping the toggle is a silent no-op.

## Changes

- **Provider editor**: Auto Compat description now states it clamps thinking levels and drops unsupported fields via the mapped pi-ai model, and that it does nothing unless the pi-ai Provider and per-model Model IDs are set. The switch is disabled when no pi-ai Provider is selected.
- **Model editor**: description now references the mapped pi-ai Model ID and the Auto Compat requirement. The checkbox is disabled unless the provider has a pi-ai Provider selected (deliberately not gated on the individual Model ID, so it can be enabled ahead of per-model mapping).
- **ModelCard**: threads the existing `piAiProvider` prop through to `ModelAdvanced`.

## Testing

- Verified all four toggle states in a real browser against the local dev stack: provider toggle disabled (no pi-ai provider) and enabled (provider set); model checkbox disabled (no provider) and enabled (provider set, including with the Model ID cleared to "none", unsaved).
- `bun run lint:check`, `format:check`, and `typecheck` all clean.

## Notes

- Motivating case: staging `google-s/gemini-3.8-flash` was mapped to pi-ai model `gemini-flash-latest` (which allows `minimal`), so an invalid thinking level passed through unclamped and Google 400'd. The staging mapping and alias were fixed directly; this PR covers only the UI follow-ups.


mcowger:
Addressed review findings in e2c734db:

- **Locked-in provider toggle**: fixed — `disabled={!editingProvider.pi_ai_provider && !editingProvider.auto_compat}`, so it can always be switched off.
- **Locked-in model toggle**: fixed the same way (`disabled={!mappingReady && modelConfig.auto_compat !== true}`).
- **Tooltip on disabled input**: fixed — `title` moved to the wrapping `<label>` (only when fully disabled), since disabled inputs suppress hover in Chrome/Safari.
- **Gate model toggle on Model ID too**: deliberately not applied — enabling Auto Compat ahead of per-model mapping is intentional (provider with several models shouldn't require every Model ID before the toggle can be used); the ID only determines whether translation applies to that model at request time.

